PR #26, opened 2014-12-01
Enabled HBRT for Palmetto in palmetto.config

joshrispoli:
LiveVPD has resolved the problem with CVPD getting corrupted (not ecc good) when HBRT is enabled via palmetto.config (DISABLE_HOSTBOOT_RUNTIME).


williamspatrick:
This pull request was out of date.  Please resubmit if still required.

PR #29, opened 2014-12-13
Removed tracepp patch file and updated occ commit level

wilbryan:
Change-Id: Idf205ee0ec6e48713c62944388b89a12d2efa0fc


williamspatrick:
Already done.

PR #36, opened 2014-12-17
Integrate Firestone MRW

rjknight:
added new openpower package to support Firestone. Need to grab pnor, firestone-xml and op-build for this to be complete.


rjknight:
Moving to serverwhiz2 data processing, will resubmit once updates are complete.

PR #38, opened 2014-12-18
Removed VDDR workaround from Habanero

bofferdn:
Co-req w/ http://gfw160.aus.stglabs.ibm.com:8080/gerrit/#/c/14710/


williamspatrick:
@wghoffa - I can't merge this directly into op-build because it will break other code.  When you update op-build (with latest hostboot, hab-xml, and palm-xml) please also remove this patch.  I suspect the patch won't even apply cleanly anymore anyhow.

williamspatrick:
Fixed with #45.

PR #43, opened 2015-01-23
Add openpower_mambo_defconfig

stewart-ibm:
This config builds just skiboot and skiroot (kernel and initramfs)
that's useful for using with the POWER8 functional simulator for
testing skiroot, kernel and petitboot functionality.


williamspatrick:
This seems a little odd to put into op-build if you only want things that are in buildroot, doesn't it?

Is there any reason why we can't use one of the existing configs and just ignore some of the output from it?

stewart-ibm:
> This seems a little odd to put into op-build if you only want things
> that are in buildroot, doesn't it?

It does a little... but it's certainly handy and means that building
specific things for Mambo follows the same build procedure as for, say
palmetto.

> Is there any reason why we can't use one of the existing configs and
> just ignore some of the output from it?

You can... but you end up building things you don't need.

I'm also likely going to put in a new kernel config for mambo to at
least exclude RAID as doing the computation as to which instructions are
best for doing RAID calculation is really slow with mambo (adds about a
minute to boot time).

PR #45, opened 2015-01-30
Update Various Package Levels for Habanero

wghoffa:
Updates for buildroot, Hostboot, OCC, Palmetto-XML, and Habanero-XML packages.

### DIFF
--- a/openpower/package/habanero-xml/habanero-xml.mk
+++ b/openpower/package/habanero-xml/habanero-xml.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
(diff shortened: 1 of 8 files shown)


wghoffa:
@jk-ozlabs - I think I mispoke on the phone. You can see my change here. It is an update to the buildroot commit level. It was more than just the skiboot update. Let me know if I still did this wrong though as I don't think I fully understand the whole picture yet.

jk-ozlabs:
Ah, that looks good to me. As long as we're pulling all of the buildroot updates, we're fine.

PR #51, opened 2015-02-13
Update hostboot commit number

ploetzma:
Updates need to update hostboot commit number to 44fde6512e834ea8f5071299ca99b35973c0f979


openpower/package/palmetto-xml/palmetto-xml.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-PALMETTO_XML_VERSION = a3fc5ca767f0447a80442bbed70eec16f67fc7f4
+PALMETTO_XML_VERSION = 283937a8237092addf07b7ab4c436337bab89e37
```
williamspatrick: Usually we should wait on these until the palmetto-xml is actually merged in so you can point at HEAD.  In this case open-power/palmetto-xml#283937 is just the old HEAD plus your changes, so it doesn't matter, but if there were other commits merged into palmetto-xml also it might.  (same goes for habanero-xml obviously)

ploetzma:
Ah I see:

commit be6c03b8498c53279e3b2e4e32d8f0c795aee940 2 parents 199f8db + 283937a

I will remember this next time.

PR #55, opened 2015-02-17
op-build pull request to support BIOS infrastructure

bofferdn:
Pre-req of "Support BIOS framework" commit @ gerrit review 15583


ozbenh:
On Mon, 2015-02-16 at 21:31 -0800, Nick Bofferding wrote:
> Pre-req of "Support BIOS framework" commit @ gerrit review 15583

Why on earth do you call that stuff "BIOS" ? :-)

Cheers,
Ben.

bofferdn:
Changed two files to get the build dependencies ordered right; BIOS because that is the term that was used in every design discussion I've been privy to, until very recently.  What should the official term be?

bofferdn:
During the reorder, didn't seem to like one of the config dependent names when not unique, so renamed.  Pull request should be complete now.

PR #58, opened 2015-02-19
Add openPOWER version string support

scprek:
See VERSION.readme for more info


scprek:
Dependent on https://github.com/open-power/pnor/pull/24

openpower/package/pkg-versions.mk
```diff
@@ -0,0 +1,137 @@
...
+
+$(2)_VERSION_FILE = $$(OPENPOWER_VERSION_DIR)/$(1).version.txt
+ALL_SUBPACKAGE_VERSIONS += $$($(2)_VERSION_FILE)
+
+### Create subpackage patch file
+define $(2)_OPENPOWER_PATCH_FILE
+
+mkdir -p "$$(OPENPOWER_VERSION_DIR)"
+
+cd "$$(PKGDIR)"; if ls *.patch; then sha512sum *.patch | sha512sum | cut -c 1-7 | \
```
williamspatrick: We also need to look for patches in the BR2_GLOBAL_PATCH_DIR location.

openpower/package/pkg-versions.mk
```diff
@@ -0,0 +1,137 @@
...
+
+# Add appropriate templates to hooks
+$(2)_POST_PATCH_HOOKS += $(2)_OPENPOWER_PATCH_FILE
+$(2)_PRE_BUILD_HOOKS += $(2)_OPENPOWER_VERSION_FILE
+
+# Create rule to print subpackage version
+$(1)-version:
+	@if [ -f $$($(2)_VERSION_FILE) ]; then \
+		cat $$($(2)_VERSION_FILE); \
+		else echo "Version file does not exist, recompile needed"; fi
```
williamspatrick: Why can't we run $(2)_OPENPOWER_PATCH_FILE in this case?

openpower/package/pkg-versions.mk
```diff
@@ -0,0 +1,137 @@
...
+
+endef ###
+
+$$(UPPER_CASE_PKG)_PRE_BUILD_HOOKS += $$(UPPER_CASE_PKG)_OPENPOWER_VERSION_FILE
+
+# Rule to print out pnor version
+$(1)-version:
+	@if [ -f $$($$(UPPER_CASE_PKG)_VERSION_FILE) ]; then \
+		cat $$($$(UPPER_CASE_PKG)_VERSION_FILE); \
+		else echo "Version file does not exist, recompile needed"; fi
```
williamspatrick: Can we do something here as well to generate this?  Maybe is a little trickier because we need some kind of foreach on ALL_SUBPACKAGE_VERSIONS

openpower/package/pkg-versions.mk
scprek: Do you mean $(2)_OPENPOWER_VERSION_FILE?

williamspatrick: Yes.  OPENPOWER_VERSION_FILE.

scprek: Now that I can compile again, I'm looking into this more.
1. I don't understand how calling $(2)_OPENPOWER_VERSION_FILE gets me what I want as is? Wouldn't I need to add a cat of $(2)_VERSION_FILE somewhere in the template?
2. Also Is this the syntax? or is it call? $(1)-version:»··»···$$($(2)_OPENPOWER_VERSION_FILE)
I've tried a lot of variations and none have worked.

openpower/package/pkg-versions.mk
scprek: Is this a separate patch sha we want? Or do we want some way of combining $$(PKG) and BR2_GLOBAL_PATCH_DIR sha's together?

williamspatrick: Instead of just $(PKGDIR)/*.patch we also need $(BR2_GLOBAL_PATCH_DIR)/$(2)/*.patch.  Two globs instead of just one.

openpower/package/pkg-versions.mk
williamspatrick: I think it would be something like this:

if [ -n $$($(2)_VERSION_FILE) ]; then $$(($2)_OPENPOWER_VERSION_FILE); fi
cat $$($(2)_VERSION_FILE)  # file always exists now.

Since $(2)_OPENPOWER_VERSION_FILE is a set of shell commands, I don't think it would be a $(call ...)

openpower/package/pkg-versions.mk
scprek: I cannot figure out why there are extra double quotes in BR2_GLOBAL_PATCH_DIR
 cd $$(BR2_GLOBAL_PATCH_DIR)/$(1); ls;
and  cd "$$(BR2_GLOBAL_PATCH_DIR)/$(1)"; ls; 
give the below result
cd "" " /esw/san2/smcprek/open-power/forks/op-build/openpower/custom/patches "/hostboot; ls;

williamspatrick: 1) BR2_GLOBAL_PATCH_DIR is a list.   So you're going to need to have a
foreach.  Do you really need to 'cd' into the BR2_GLOBAL_PATCH_DIR though?
Why can't you just 'ls *.patch $$(foreach
path,$$(BR2_GLOBAL_PATCH_DIR),$path/$(1))'

2) There might be an issue with how I defined BR2_GLOBAL_PATCH_DIR.  Check
out pkg-generic.mk in buildroot and see how they use it.  I recently added
a BR2_GLOBAL_PATCH_DIR+=... in openpower/external.mk and maybe I did it
wrong.

scprek: Yeah there could be something wrong with the external.mk change, I'm trying things now. I originally did the cd to just have an example of what the variable looks like.
I couldn't get ls *.patch $path/$(1) to work without cd'ing into the dir first, so I'm trying find now
$$(foreach path, $$(BR2_GLOBAL_PATCH_DIR),find $$(path)/$(1) -name "*.patch" -printf '%f\n')

scprek: Looks like external.mk needs a change
The spaces before and after the variable caused problems
BR2_GLOBAL_PATCH_DIR += " $(BR2_EXTERNAL)/custom/patches "
Changed to
BR2_GLOBAL_PATCH_DIR += "$(BR2_EXTERNAL)/custom/patches"

openpower/package/pkg-versions.mk
scprek: I'm having issues with this, I keep getting syntax errors when I put $$($(2)_OPENPOWER_VERSION_FILE) in an if statement, so I decided to make it more basic to see what's happeneing.

$(1)-version: $$($(2)_OPENPOWER_VERSION_FILE) 

make: Entering directory `/esw/san2/smcprek/open-power/forks/op-build/buildroot'
make: *** No rule to make target `mkdir', needed by `hostboot-version'.  Stop.
make: Leaving directory `/esw/san2/smcprek/open-power/forks/op-build/buildroot'

mkdir is the first command in define $(2)_OPENPOWER_VERSION_FILE
Also I tried a basic rule define STEVE echo "steve" and I got a similar error no rule to make target echo.

scprek: Well I got to work and realized I put that on the dependency line vs below which seems to work.
$(1)-version: 
            $$($(2)_OPENPOWER_VERSION_FILE)

scprek:
Even though this is based on master-next, the push updated this pull request.
I will close this pull request as https://github.com/open-power/op-build/pull/93 has been created.

PR #69, opened 2015-02-28
New OCC version for op-build.

sbroylesibm:
New OCC version for op-build.  Update to occ.mk.


geissonator:
I did a quick sniff on hab19, looks good.

PR #73, opened 2015-03-01
Enable SBE Updates

geissonator:


geissonator:
Note this requires IPMI to be enabled (which is the default) since it utilizes the watchdog to force the automatic update.  If you don't have IPMI enabled then you'll probably just have to manually power down and back up again.

PR #79, opened 2015-03-03
Create 'custom' subdirectory.

williamspatrick:
This directory is a place for users to add custom packages
    and patches outside of the rest of the openpower package
    tree.  This allows them to more easily maintain a fork of
    op-build and enhance it with their own packages.

### DIFF
--- a/openpower/Config.in
+++ b/openpower/Config.in
@@ -8,4 +8,8 @@ source "$BR2_EXTERNAL/package/Config.in"
 endif
 endmenu
 
+menu "Custom Packages"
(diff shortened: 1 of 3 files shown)


dcrowell77:
Seems like a good idea to me.

geissonator:
@williamspatrick , can you provide an example of how we'd use this?  Should we wait for the phase 4 tag before merging or low risk?

williamspatrick:
Added README file.

williamspatrick:
Merged to master-next.

PR #83, opened 2015-03-04
buildroot: Update to 9355c73e

jk-ozlabs:
This change includes a set of fixes for skiboot on Palmetto & Habanero:

 - IPMI watchdog disable fixes
 - i2c interface sync with HBRT
 - PNOR partition offset
 - Use Firmware FRU ID from device tree
 - Use NULL for FSP-only hservices callbacks


jk-ozlabs:
.. and the buildroot tree is up-to-date this time :)

PR #86, opened 2015-03-05
Build bootloader kernel seperate from skiboot

shenki:
We want the kernel to be placed in it's own BOOTKERNEL partition on the
flash. In that case we don't want it linked into the skiboot image,
instead it should be incorporated independently in to the PNOR image.

This is made the default for Palmetto and Habanero machines.


williamspatrick:
+1

williamspatrick:
Merged into master-next.

PR #88, opened 2015-03-06
Add CAPI ucode support

mikey:
This adds building the CAPI ucode into the palmetto and habenero pnor images.

Uses recently merged CAPI support in the pnor git tree.

Mikey


williamspatrick:
Merged into master-next.

PR #93, opened 2015-03-10
Add openPOWER version string support

scprek:
See VERSION.readme for more info


scprek:
Dependent on https://github.com/open-power/pnor/pull/24

scprek:
If we want each subpackage version to be the most up to date, I feel that we should run the rules everytime you do op-build hostboot-version, vs checking if the file exists first.

stewart-ibm:
Do we want to add in the SHA of the *binary* included into the image? Since we may be able to flash individual partitions, the global version strings may end up being incorrect. Although this situation is most likely in the lab, so perhaps it doesn't matter - and we just say "don't do that" for anyone in the field?

For hostboot and skiboot it'd be possible for them to programmaticaly communicate their version numbers up and out to Linux so that a user could find out what they're running. I wonder if there's any reason we couldn't also do this with OCC? We probably also want version to contain something about the skiroot (kernel+initramfs+petitboot) environment.

FWIW - skiboot versions will be X.Y rather than just the hash.

scprek:
Note from VERSION.readme
$(version) = git tag or first 7 characters of commit

Therefore skiboot would show up as skiboot-ver1.0-.... vs skiboot-ac5de45

williamspatrick:
How could you ever get back to a file version from the hash of the binary?   Even two different compiles of the exact same code could result in a different hash.                                                                                                                                                                                                                                          From: Stewart SmithSent: Thursday, March 12, 2015 6:50 AMTo: open-power/op-buildReply To: open-power/op-buildSubject: Re: [op-build] Add openPOWER version string support (#93)Do we want to add in the SHA of the binary included into the image? Since we may be able to flash individual partitions, the global version strings may end up being incorrect. Although this situation is most likely in the lab, so perhaps it doesn't matter - and we just say "don't do that" for anyone in the field?

For hostboot and skiboot it'd be possible for them to programmaticaly communicate their version numbers up and out to Linux so that a user could find out what they're running. I wonder if there's any reason we couldn't also do this with OCC? We probably also want version to contain something about the skiroot (kernel+initramfs+petitboot) environment.

FWIW - skiboot versions will be X.Y rather than just the hash.

—Reply to this email directly or view it on GitHub.

stewart-ibm:
> How could you ever get back to a file version from the hash of the
> binary?   Even two different compiles of the exact same code could
> result in a different hash.

It'd be a rather crude way of detecting if individual partitions have
been flashed. Possibly not a *good* idea.

Actually... fully repeatable builds really would be a good idea for
us. Something to try in our infinite spare time :)

scprek:
removed some trailing whitespace, noticed when making a patch for this change.

scprek:
Removed common-p8-xml and added capp-ucode

PR #131, opened 2015-04-08
Update firestone XML and patches needed for Firestone

ploetzma:
Two patches and a firestone XML update needed for Firestone.

### DIFF
--- a/openpower/package/firestone-xml/firestone.mk
+++ b/openpower/package/firestone-xml/firestone.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-FIRESTONE_XML_VERSION ?= 8023db709f2ed5ee4c481fcb7eff68a30248e0ae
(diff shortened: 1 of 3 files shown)


ploetzma:
Hold off on this a second - I think one of the patches does not like Palmetto.

williamspatrick:
The vpd one seemed to have some problems on another system. ‎ That would be my first suspect.                                                                                                                                                                                                                                          From: Matt PloetzSent: Thursday, April 9, 2015 9:42 AMTo: open-power/op-buildReply To: open-power/op-buildSubject: Re: [op-build] Update firestone XML and patches needed for Firestone (#131)Hold off on this a second - I think one of the patches does not like Palmetto. 

—Reply to this email directly or view it on GitHub.

ploetzma:
Yes. The VPD commit will not work with out a new target types from Bill Schwartz. 
My best idea from here is to separate the VPD fix from the TX_MSBSWAP commit and make a patch for only the TX_MSBSWAP fix.

PR #137, opened 2015-04-14
OCC firmware update for Habanero SoftLayer delivery.

sbroylesibm:
OCC build op_occ_150415a

### DIFF
--- a/openpower/package/occ/occ.mk
+++ b/openpower/package/occ/occ.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


williamspatrick:
Do we need a patch for a specific problem for op-build 1.1.1?  Since this is against master-next it is going to end up in op-build 1.2.

williamspatrick:
Merged into master-next for 1.2.

sbroylesibm:
Patrick I confirmed that we do not need to patch the v1.1 release with any of the content pushed to occ master today.

PR #141, opened 2015-04-21
Update capp-ucode to lastest repo

mikey:
Fixes issue for build systems where /bin/sh != /bin/bash


shenki:
@williamspatrick We weren't sure which branch this should be against. Ideally we would have the fix in all branches, so that our CI systems will continue to work.

PR #157, opened 2015-04-29
Fixed palmetto strlen exception

e-liner:


williamspatrick:
Was anyone asking for this early?  I'd think we could wait until 1.3.

williamspatrick:
Lets handle this as a hostboot update.

PR #158, opened 2015-04-30
Update skiboot to skiboot-5.0 release (rather than -rc3)

stewart-ibm:

It should be noted that master has commits that master-next doesn't.


williamspatrick:
Master has 1 commit which is a merge commit ahead of master-next. Master-next has real content ahead of master but not the other way around.

williamspatrick:
Deprecated in favor of issue #166.

PR #161, opened 2015-04-30
Fix handling of ECC procted partitions during runtime

ploetzma:


williamspatrick:
Manually merged to master.

PR #166, opened 2015-05-01
OPAL updates for Habanero

jk-ozlabs:
These changes implement a set of fixes for Habanero.

On the OPAL side, we update to skiboot-5.0.1, which means we no longer need the out-of-tree patch for IPOLL masking, and prevent a hang on XSCOM failures.

We bring in a few linux changes, mostly for preventing invalid error messages in non-FSP code

We also bring in a petitboot fix for the OS Boot sensor.

PR #169, opened 2015-05-04
Switch to a little-endian kernel & userspace

jk-ozlabs:
This change (on top of my pending pull request https://github.com/open-power/op-build/pull/166) switches the kernel & userspace to little-endian.

I've tested the kernel & petitboot parts, but it seems there's still an issue with either hostboot or OCC, a I just get a bunch of OCC failure logs.


jk-ozlabs:
@williamspatrick as discussed - this may break booting, but at least we can work out what needs fixing.

williamspatrick:
I haven't had a chance to play with this yet.

bwmashak:
Suppport for PMC adapters in IBM version of Habanero is dependent upon this LE kernel and userspace.  Also, the PMC driver needs to be included in an additional pull request to allow for user to test this feature.  Can this be part of v1.3 of op-build, so boot testing from devices downstream of PMC can begin?

williamspatrick:
v1.3 is already made. This can get into 1.4.

Jeremy has a pull request to enable LE builds but he said it broke something in hostboot or occ. I have not had a chance to test it out and see what is broken.

The kernel driver should be sent to Jeremy for inclusion into the kernel tree they maintain for Op-build.

bwmashak:
Kernel driver sent to Jeremy.  What is the schedule for v1.4?

williamspatrick:
https://github.com/open-power/op-build/milestones/

williamspatrick:
Jeremy, 

The problems with OCC seem to have been resolved by decreasing the optimization to -O1.  We had to reduce the optimization in Hostboot from -O3 to -O2 previously to get around these problems.

I'm not seeing these optimization problems with our own Hostboot cross-compiler, which is 4.9.2.  Do you think we could upgrade this pull from 4.9.0 to 4.9.2 and see if these issues go away?

ozbenh:
On Thu, 2015-07-02 at 09:17 -0700, Patrick Williams wrote:
> Jeremy, 
> 
> The problems with OCC seem to have been resolved by decreasing the
> optimization to -O1. We had to reduce the optimization in Hostboot
> from -O3 to -O2 previously to get around these problems.
> 
> I'm not seeing these optimization problems with our own Hostboot
> cross-compiler, which is 4.9.2. Do you think we could upgrade this
> pull from 4.9.0 to 4.9.2 and see if these issues go away?

Moving away from a .0 gcc is always a good idea :)

Cheers,
Ben.

jk-ozlabs:
OK, it should be fine to update to 4.9.2. There might be a bit of buildroot-hackery involved, but that should be fairly straightforward. I'll take a look - I'm travelling at the moment, so i'll likely be a little slow at that -  feel free to jump in before me :)

fmkassem:
Moving to 4.9.2 won't resolve this issue.
I was able to narrow it down to a specific optimization flag that is causing OS traps to be introduced early in the code.   The flag is new to gcc 4.9 and wasn't part of the gcc4.8 or earlier.  
The culprit flag is -fisolate-erroneous-paths-dereference.   I'll be looking into it and whether we can manipulate the code to not cause these default "optimization" to happen.
One workaround is to add the -fno-isolate-erroneous-paths-dereference to the make file for OCC that will disable this optimization that is enabled via -Os.

ozbenh:
On Tue, 2015-07-07 at 16:14 -0700, Fadi Kassem wrote:
> Moving to 4.9.2 won't resolve this issue.
> I was able to narrow it down to a specific optimization flag that is
> causing OS traps to be introduced early in the code. The flag is new
> to gcc 4.9 and wasn't part of the gcc4.8 or earlier.

I have been bitten in the past with gcc inserting trap
instructions :-( I had cases where I was dereferencing NULL (correctly !
I really had something at physical 0 !) and it was trapping on it...

However, it might be interesting to figure out whether in this case gcc
is flagging some real issue in the code..

fmkassem:
That's true Ben, in our case too we have a pointer to address 0x00000000 that maps to real memory.  But the compiler sees it as a NULL pointer, thus the trap.

To fix this I lowered the Optimization level for that function alone to 1.
Code should be pushed up to github today.

stewart-ibm:
I wonder if we should file a GCC bug that -fno-isolate-erroneous-paths-dereference should be set when -fstandalone is set? Odds are a bunch of other low level projects will hit this at some point.

ozbenh:
On Thu, 2015-07-16 at 14:47 -0700, Fadi Kassem wrote:
> That's true Ben, in our case too we have a pointer to address
> 0x00000000 that maps to real memory. But the compiler sees it as a
> NULL pointer, thus the trap.
> 
> To fix this I lowered the Optimization level for that function alone
> to 1.
> Code should be pushed up to github today.

There are other ways to fix it but to be honest I can't remember off the
top of my head...

fmkassem:
Yes, we have ideas for other fix options, but this seems to be the most appropriate given the time and the various levels of GCC that could be used to compile this code.    Also, the reduction in optimization is very minimal and has a very slight effect on the code.

jk-ozlabs:
So are we okay to merge this, with an addition of CFLAGS to the occ build process?

jk-ozlabs:
(or are those changes already in OCC?, and this pull request can be merged as-is?)

fmkassem:
The necessary changes have been made in OCC commit 0726e69, therefore no changes are needed in this pull.

williamspatrick:
We need to test having this pull request added to master-next again.  The OCC changes are in master-next now.

williamspatrick:
Jay tested this after the occ change and it looks good. 

@maploetz - can you make similar changes for garrison when you get a chance?  Garrison was added after this pull request was created.

PR #172, opened 2015-05-14
Make pkg-versions.mk handle versions/tags with package name

stewart-ibm:
Some SCM tags/versions can include the package name, in this event, we want
to exclude the package name from the generated version that ends up
in the VERSION partition.

This patch simply uses some sed to rip out 'package-' from anything from
git.


williamspatrick:
@scprek - Please review.

PR #176, opened 2015-05-21
Increase UART timeout to 10s

bofferdn:
Up UART timeout to 10s to compensate for loaded BMC during auto boot


williamspatrick:
Manually merged into master-next.

PR #180, opened 2015-06-02
Enable Hostboot to Detect Hardware Changes

wghoffa:
@dcrowell77 - I think this is what was needed for the GARD issue?

@ploetzma - FYI on the op-build change.


dcrowell77:
Yep, that should be the change.

williamspatrick:
Needs to be against master-next.

PR #182, opened 2015-06-04
Enable OCC for Firestone

cjcain:
This can be pulled in but for dual OCC functionality you also need to pull in http://gfw160.aus.stglabs.ibm.com:8080/gerrit/#/c/18179/


williamspatrick:
Manually applied to master-next along with the co-req commit from Doug and XML update.

PR #195, opened 2015-06-24
op-build hostboot package update June 23th

jazurin:
updated hostboot version to 4d1c59f09b6b0ab9236aa42694cb2e42cbe8d3eb
Tested on palmetto, habanero and firestone.


jazurin:
Note that this is an incremental update.  There are newer changes on gerrit/master that are not included here.  We are debugging an IPL issue introduced by those commits.  Once the issue is fixed I'll create another pull request to include those changes.

Issue #216, opened 2015-07-31
Unable to build master-next


nacc:
Attempting to build garrison_defconfig, I get:

>>> linux d0ff068a9962ba96d4e6ddfba2a4e9fd8bc9a763 Building
PATH="/home/nacc/op-build/output/host/bin:/home/nacc/op-build/output/host/sbin:/home/nacc/op-build/output/host/usr/bin:/home/nacc/op-build/output/host/usr/sbin:/home/nacc/bin:/home/nacc/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games" /usr/bin/make -j9 HOSTCC="/usr/bin/gcc" HOSTCFLAGS="" ARCH=powerpc INSTALL_MOD_PATH=/home/nacc/op-build/output/target CROSS_COMPILE=" /home/nacc/op-build/output/host/usr/bin/powerpc64-buildroot-linux-gnu-" DEPMOD=/home/nacc/op-build/output/host/sbin/depmod -C /home/nacc/op-build/output/build/linux-d0ff068a9962ba96d4e6ddfba2a4e9fd8bc9a763 zImage.epapr
make[1]: Entering directory '/home/nacc/op-build/output/build/linux-d0ff068a9962ba96d4e6ddfba2a4e9fd8bc9a763'
scripts/kconfig/conf --silentoldconfig Kconfig
  CHK     include/config/kernel.release
  WRAP    arch/powerpc/include/generated/asm/clkdev.h
  WRAP    arch/powerpc/include/generated/asm/hash.h
  WRAP    arch/powerpc/include/generated/asm/irq_work.h
  WRAP    arch/powerpc/include/generated/asm/mcs_spinlock.h
  WRAP    arch/powerpc/include/generated/asm/preempt.h
  WRAP    arch/powerpc/include/generated/asm/scatterlist.h
  WRAP    arch/powerpc/include/generated/asm/rwsem.h
  WRAP    arch/powerpc/include/generated/asm/trace_clock.h
  WRAP    arch/powerpc/include/generated/asm/vtime.h
  CHK     include/generated/uapi/linux/version.h
  UPD     include/generated/uapi/linux/version.h
  UPD     include/config/kernel.release
  HOSTCC  scripts/basic/bin2c
  CHK     include/generated/utsrelease.h
  UPD     include/generated/utsrelease.h
  CC      kernel/bounds.s
[...]

PR #217, opened 2015-07-31
configs: Update garrison to use a le build

jk-ozlabs:
Fixes https://github.com/open-power/op-build/issues/216


williamspatrick:
Jeremy,

I understand we should move all platforms to LE but why do we have to?  Can you also fix whatever compile failure we are running into when we do compile in BE mode. I'm not sure if all the partners want to us LE or what their timeline for transitioning is.

PR #226, opened 2015-08-13
op-build update August 12th

jazurin:
Updated hostboot version. Tested on habanero, palmetto, and firestone
HW.


openpower/configs/hostboot/firestone.config
```diff
@@ -30,7 +30,9 @@ set GPIODD
 set PALMETTO_VDDR
 
 # disable sbe updates
-set NO_SBE_UPDATES
+#set NO_SBE_UPDATES
```
williamspatrick: We should remove this rather than comment it out.

PR #230, opened 2015-08-17
Update kernel to include kexec fixes

sammj:
Depends on https://github.com/open-power/linux/pull/2 ("Include kexec endianess fixes for LE builds")


shenki:
@williamspatrick this is good to go into master-next. We need so our LE petitboot can to boot BE kernels

sammj:
Will resend - hashes have changes since being pulled in

PR #240, opened 2015-08-18
Reduce kernel size

stewart-ibm:
This series disables a bunch of kernel config options that end up saving us several hundred KB in the petitboot kernel.

This should likely aim for 1.7 rather than 1.6


openpower/configs/linux/skiroot_defconfig
```diff
@@ -1,5 +1,7 @@
 CONFIG_PPC64=y
 CONFIG_ALTIVEC=y
+CONFIG_PPC_POWERNV_RTAS=n
+CONFIG_PPC_PSERIES=n
```
shenki: Does this affect us being able to test in Qemu?

stewart-ibm: Only if you don't use @ozbenh powernv qemu. Which we should probably move to. Only used in petitboot-tests atm.

stewart-ibm:
This last commit shaves about 900k off zImage.epapr size.

PR #252, opened 2015-08-24
IOP settings and Garrison Opal Model

Erich-Hauptli:
Modified Garrison_XML_Version to point to latest garrison.xml contains multiple fixes

-APSS GPU update
-IOP updates
-OPAL pointer switched from Firestone to Garrison


williamspatrick:
Garrison appears to already point at this level.

PR #255, opened 2015-08-28
Update buildroot to 2015.05 base

stewart-ibm:
Adds petitboot to openpower overlay rather rather than in buildroot branch.

This pull request is a RFC for the moment and some review is likely needed and very much appreciated.

This uses the buildroot referenced in https://github.com/open-power/buildroot/pull/8

It also moves the petitboot package into the openpower overlay rather than as an additional package in buildroot.

This BuildsForMe(TM) and BootsForMe(TM).


stewart-ibm:
We are good to merge this in - everything seems to check out as good. We have some additional pull reqs for 1.7, but they'll be on top of this.

buildroot changes are in a open-power/buildroot branch (rebased our patches on upstream buildroot - and we can flip open-power/buildroot master over to that branch once we merge this).

PR #264, opened 2015-09-11
Set kernel version to 4.2

shenki:
We switch from using our own tree, to building with upstream tarballs.
Currently we still require a small number of out of tree patches; this
list will shrink as we get the last of the patches upstream, to the
point where will just set the version.

The patches are generated from the tree at
https://github.com/open-power/linux still, and imported into the
buildroot overlay.


stewart-ibm:
depends on https://github.com/open-power/op-build/pull/255 (buildroot update)

This fixes several bugs for us as well as including some needed patches for still-to-come raid driver.

PR #266, opened 2015-09-12
op-build update Sep 11th

jazurin:
Updated versions: hostboot, habanero-xml
Tested on palmetto, habanero, and firestone HW.  Garrison build only

### DIFF
--- a/openpower/package/habanero-xml/habanero-xml.mk
+++ b/openpower/package/habanero-xml/habanero-xml.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-HABANERO_XML_VERSION ?= 3ad2b324330987755584d7d0aa1d19f18c9e225e
+HABANERO_XML_VERSION ?= f73630d727dcae2496c3f69e47b6b83ff96aab0f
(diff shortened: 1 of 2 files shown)


jazurin:
@williamspatrick I tried removing https://github.com/open-power/op-build/blob/master/openpower/package/hostboot/hostboot-0002-GCC-4.9-Make-compiler-use-ELFv1-ABI-and-use-O2.patch but could not get the systems to boot. Didn't see SOL output and could not get a hostboot dump to debug.  I did not look into it further and put the patch back.

williamspatrick:
Thanks for trying.

On Fri, Sep 11, 2015 at 08:05:42PM -0700, Jay Azurin wrote:
> @williamspatrick I tried removing https://github.com/open-power/op-build/blob/master/openpower/package/hostboot/hostboot-0002-GCC-4.9-Make-compiler-use-ELFv1-ABI-and-use-O2.patch but could not get the systems to boot. Didn't see SOL output and could not get a hostboot dump to debug.  I did not look into it further and put the patch back.
> 
> ---
> https://github.com/open-power/op-build/pull/266#issuecomment-139705342

-- 
Patrick Williams

PR #267, opened 2015-09-14
Add dependencies for software RAID

sammj:
Petitboot can support filesystems on software RAID arrays, but needs
mdadm and the associated udev rules to assemble the device.

Signed-off-by: Samuel Mendoza-Jonas <sam.mj@au1.ibm.com>


williamspatrick:
Is there a reason only the palmetto_defconfig is updated?

sammj:
Just me not looking closely enough at the commit! Updated to properly include the rest of the platforms.

PR #276, opened 2015-09-24
Enable translations for petitboot

jk-ozlabs:
This change adds translations for petitboot. We already have the message
catalogues for a bunch of languages, so enable building these into the
final image.

We enable these for firestone and habanero, but there's no reason we
couldn't do this for other languages too.


williamspatrick:
Looks fine to me.  @jk-ozlabs - Can you have someone on your team check it out.

stewart-ibm:
looks good from here

PR #281, opened 2015-09-29
linux: Update to v4.2.1-openpower1

shenki:
This brings in the 4.2.1 stable release, and the accraid RAID driver for
PMC-Sierra RAID  on Habanero.

The RAID driver is submitted upstream to the linux-scsi mailing list,
but is not yet in a tree.

The stable release brings in a few commits that are of interest to
OpenPower platforms:

 94d718d024e xfs: return errors from partial I/O failures to files
 2937b784b19 xfs: Fix file type directory corruption for btree directories
 95310b79776 libxfs: readahead of dir3 data blocks should use the read verifier
 0a3b19cef4f sched: Fix cpu_active_mask/cpu_online_mask race


shenki:
Stewart, I regenerated our skiboot_defconfig. Can you double check the changes; the only functional change should be the addition of the accraid driver (CONFIG_SCSI_AACRAID=y)

stewart-ibm:
I do, in fact, have a regenerated defconfig laying around as this was something I was going to push in as well (precisely so this kind of patch was easy).

The diff is:
--- a/openpower/configs/linux/skiroot_defconfig
+++ b/openpower/configs/linux/skiroot_defconfig
@@ -79,6 +79,7 @@ CONFIG_SCSI_CXGB3_ISCSI=y
 CONFIG_SCSI_CXGB4_ISCSI=y
 CONFIG_SCSI_BNX2_ISCSI=y
 CONFIG_BE2ISCSI=y
+CONFIG_SCSI_AACRAID=y
 CONFIG_SCSI_MPT2SAS=y
 CONFIG_SCSI_SYM53C8XX_2=y
 CONFIG_SCSI_SYM53C8XX_DMA_ADDRESSING_MODE=0


i.e. *exactly* what we want, just enabling AACRAID.

williamspatrick:
Is someone updating this pull request without the skiboot_defconfig update then?  Should we just include AACRAID in the defconfig?

shenki:
@williamspatrick that's what we've done. Stewart and I were just confirming that there were no other changes present.

Lets merge this in.

dougmill-ibm:
I've been having trouble building because of the patches in 
openpower/linux. These patches seem to be already applied to the 
4.2.1-openpower1 kernel. When I delete the patch files, the build works.

Is this something that was missed, or am I building incorrectly?

Thanks,
Doug

shenki:
Hey Doug,

When building the firmware, the upstream kernel tarball is used (in this case 4.2.1), and the patches in the ```openpower/linux``` directory are applied to that tarball.

In order to generate the patches that we stick in ```openpower/linux```, I maintain a kernel tree that contains the commits on top of a stable release. Those commits are split out into patches and committed to op-build.

So if you want to do some testing, you can check out the tag from the https://github.com/open-power/linux repository, and build a tag such as https://github.com/open-power/linux/releases/tag/v4.2.1-openpower1 without applying the patches.

Or if you would prefer you can use op-build to download and build the firmware, which will apply the patches to an upstream tarball for you.

Let me know if you need any further clarification.

PR #287, opened 2015-10-02
Remove dependency on realpath in post-build rootfs setup.

dougmill-ibm:
Ran into problems on other build systems where the post-build rootfs fix-up script was failing to run when "realpath" was not available. It should not be necessary to perform the desired check anyway.
A secondary issue, not addressed here, is that the build did not detect the failure.

### DIFF
--- a/openpower/scripts/fixup-target-var
+++ b/openpower/scripts/fixup-target-var
@@ -5,8 +5,7 @@
 find $TARGET_DIR/var/ -type l |
 while read path
 do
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
Looks good to me. Patrick - can you pull this?

williamspatrick:
Manually merged since this pull request came against master instead of master-next.  This did not make the v1.7 tagging but will make v1.8.

PR #296, opened 2015-10-20
linux: update to v4.2.3-openpower1

shenki:
- Dropped 'net/bnx2x: Add shutdown handler' as this is upstream
 - Added 'fbcon: initialize blink interval before calling fb_set_par'
 - Rebased on v4.2.3 stable release


shenki:
Boot tested on palmetto

PR #300, opened 2015-10-27
linux: update to v4.2.4-openpower1

shenki:
- Rebased on upstream stable release v4.2.4
 - Upstream commit "8c24d6d7b09d usb: xhci: stop everything on the first
   call to xhci_stop" means that our patch "xhci: do not halt the
   secondary HCD" is no longer required


shenki:
Boot tested to host OS on Palmetto

PR #302, opened 2015-10-29
Bump buildroot to 2015.08.1

stewart-ibm:
In the process, sync our openpower/device_table.txt with upstream
buildroot.

See https://github.com/open-power/buildroot/pull/15 for details of what changed in buildroot.


williamspatrick:
The buildroot pull request is still open.  When should we merge this?

stewart-ibm:
I think @shenki branch is probably the one we want to merge instead, squashes a few of the patches so we have a smaller delta from upstream. I think it would be okay to merge pre op-build 1.8, although that may depend on when you're planning on tagging it :)

stewart-ibm:
Since we've merged 2015.11, we don't need this anymore.

Issue #307, opened 2015-11-04
xhci bug causes loss of usb after kexec 


jazurin:
This has been seen 2 times during OP HW CI since November 2nd (2 times out of 24 boots).  We are booting using op-build master-next code on a habanero. 
It looks like an exception causes a shutdown. 
For reference here are the SOL logs of each failed test run:
https://gist.github.com/jazurin/82b49415d88842a1e105
https://gist.github.com/jazurin/de1893a90d950a3997fd

```
The system is going down NOW!

Sent SIGTERM to all processes

Sent SIGKILL to all processes
cpu 0x0: Vector: 300 (Data Access) at [c000000fd13936e0]
    pc: c0000000009cbe74: xhci_irq+0x254/0x17cc
    lr: c0000000009cbc5c: xhci_irq+0x3c/0x17cc
    sp: c000000fd1393960
   msr: 9000000000009033
   dar: 28
 dsisr: 40000000
  current = 0xc000000fc82236c0
  paca    = 0xc00000000fe80000   softe: 0    irq_happened: 0x01
    pid   = 4039, comm = kexec
enter ? for help
[c000000fd1393960] 00003fffd38f2758 (unreliable)
[c000000fd1393af0] c00000000099922c usb_hcd_irq+0x44/0x68
[c000000fd1393b10] c0000000009aa4c8 usb_hcd_pci_remove+0x44/0x140
[c000000fd1393b50] c0000000009afeac xhci_pci_remove+0x64/0x8c
[c000000fd1393b80] c0000000003d4bc0 pci_device_shutdown+0x50/0x88
[c000000fd1393bb0] c00000000059ada4 device_shutdown+0x170/0x1e8
[c000000fd1393c50] c000000000072390 kernel_restart_prepare+0x4c/0x60
[c000000fd1393c70] c0000000000b98bc kernel_kexec+0x68/0xbc
[c000000fd1393ce0] c0000000000728f4 SyS_reboot+0x1b4/0x1f4
[c000000fd1393e30] c000000000009198 system_call+0x38/0xd0
--- Exception: c01 (System Call) at 00003fffa1ce1084
SP (3fffd38f1f80) is in userspace
0:mon> 
```

jazurin:
We hit this problem again.  Any news on this issue?

Thanks.

[...]

PR #315, opened 2015-11-25
Modify XML Pointer

Erich-Hauptli:
Pick up updates to FRU IDs to match Firestone and Membuf Sensor ID updates to allow thermal algorithm to work properly.

### DIFF
--- a/openpower/package/garrison-xml/garrison.mk
+++ b/openpower/package/garrison-xml/garrison.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-GARRISON_XML_VERSION ?= 18da29a4968317e8beecd49150d04941baa36914
+GARRISON_XML_VERSION ?= 48849639151da99f5204c114eeb0af870a0fc835
 GARRISON_XML_SITE ?= $(call github,open-power,garrison-xml,$(GARRISON_XML_VERSION))
 
 GARRISON_XML_LICENSE = Apache-2.0


williamspatrick:
@Erich-Hauptli - This cannot be merged due to build failure.  Please check with @jazurin on the failure.  Are we missing a BMC or Hostboot dependency?

jazurin:
The HW CI test failed on habanero while trying to load the OS.  That is issue is being tracked here:
https://github.com/open-power/op-build/issues/307
For reference, I captured the SOL output and copied it here: https://gist.github.com/jazurin/23ffc974d6e394792de0

I will manually re-trigger the test.

Erich-Hauptli:
The test was run against a Habanero.  This patch simply updates the SHA for the Garrison-xml file.  I have no idea why you'd test this against a Habanero or why it would fail, but this patch would have nothing to do with the fail.

Erich-Hauptli:
Jay,
        I'm very confused here.  Why would you run any of this on a 
Habanero?  This change only applies to Garrison.  Please look at the 
commit as I only modified the SHA pointer for the Garrison-xml.  If 
something broke it was not due to this patch.
Thank You,
Erich Hauptli
Open Power Development Engineer
Office:    1-512-286-6566 
(T/L):      2-363-6566
Cell:       1-517-881-2927

jazurin:
The automated HW tests run on palmetto, habanero, and firestone platforms (soon on garrison).   I understand that your patch did not cause the failure.  I was simply trying to explain that the test failed due to an intermittent problem that is being looked at by the OPAL team.

PR #320, opened 2015-12-11
Update hostboot to master-p8

pragupta:


williamspatrick:
@op-jenkins test this please

williamspatrick:
I was wrong.  The branch isn't even specified here, so there is nothing for us to do.  The cronjob we already have will create a pull request to update to the latest version.

PR #323, opened 2015-12-16
Package/petitboot : Add NVME tool support

mamatha4:
Enable NVME block device and nvme tool support in petitboot
After enabling NVME device and tool in op-build this will increase the size 
of rootfs.tar by 92KB and zImage.epapr by 41KB 
Signed-off-by: Mamatha <mamatha4@linux.vnet.ibm.com>


williamspatrick:
Should we close this then?

stewart-ibm:
Yes, we can track in other pull requests or an overarching issue for full NVME support.

PR #324, opened 2015-12-22
Add hostboot build fixes patches

stewart-ibm:
This adds in current outstanding hostboot patches to fix the build on modern linux distros.


shenki:
```
Global symbol "$g_TOCEyeCatchBackup" requires explicit package name at /scratch/joel/op-build/output/host/usr/powerpc64le-buildroot-linux-gnu/sysroot/hostboot_build_images//buildpnor.pl line 518.
```

williamspatrick:
The patches we intend to merge into Hostboot have now been merged.  There is some problems with our automated tools if we merge this patchset as is because it is missing Change-Ids on each of the patches.

I poked our tool that automatically creates a pull request when the hostboot (and other) packages are updated. This pull request is therefore now deprecated by #330 and no longer needed.

PR #331, opened 2016-01-11
This patch is to enable NVME block device in petitboot

mamatha4:
Enable NVME block device and nvme tool support in petitboot
After enabling NVME device and tool in op-build this will increase the size
of rootfs.tar by 92KB and zImage.epapr by 41KB 

Signed-off-by: Mamatha <mamatha4@linux.vnet.ibm.com>


stewart-ibm:
I note that the buildroot part has been accepted in upstream buildroot now!

I've cherry-picked the bulidroot commit into our buildroot branch, and merged this patch into op-build master-next with a few minor changes.

PR #333, opened 2016-01-14
Allow for building behind at HTTP(S) proxy

deece:
This patch allows op-build to execute without a direct Internet connection.

PR #334, opened 2016-01-19
Findings from LTC

deece:
Sorry, my last pull request went to master, not master-next.

This addresses some issues we discovered at LTC:
- Incorrect license declaration
- Inconsistent fetch method for Petitboot (uses git, while others use HTTPS)
- Logic error triggering the build of the pnor update file

### DIFF
--- a/openpower/configs/firestone_defconfig
+++ b/openpower/configs/firestone_defconfig
@@ -1,7 +1,6 @@
 BR2_powerpc64le=y
 BR2_powerpc_power8=y
 BR2_OPENPOWER_PLATFORM=y
-BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL)/package"
 
 BR2_BINUTILS_EXTRA_CONFIG_OPTIONS="--enable-targets=powerpc64-linux"
 BR2_EXTRA_GCC_CONFIG_OPTIONS="--enable-targets=powerpc64-linux"
--- a/openpower/configs/firestone_defconfig
+++ b/openpower/configs/firestone_defconfig
@@ -1,6 +1,7 @@
 BR2_powerpc64le=y
 BR2_powerpc_power8=y
 BR2_OPENPOWER_PLATFORM=y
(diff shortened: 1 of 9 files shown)


stewart-ibm:
Is this just a new revision of #333 ?

This looks like it's starting to fix several things, and the title of "Findings from LTC" isn't very descriptive as to why they're grouped together.

This looks like it should be 3 series of patches:
1) pull petitboot via buildroot github infrastructure
2) fix license information
3) fix pnor building.

stewart-ibm:
Could you clean up this patch set to be a simple set of revisions against master-next and resubmit?

Thanks.

williamspatrick:
Stewart, can you just apply the change into master-next? I doubt there is anything significantly different. 

Patrick

williamspatrick:
The original change I mean and not this one. This one has merge commits that are extraneous and should not be merged anyhow.

stewart-ibm:
I merged the HTTPS proxy fixup yesterday.. I'll poke Alistair for the other changes.. I assume you're okay with the -Wno-error for Hostboot for the time being?

williamspatrick:
Any idea what error was failing to compile?  Since we have different compile flags for op-build than proprietary firmware, it is useful to know about these. I'd rather not turn off compile warnings unless there are specific issues that have been reported to the open-power/hostboot project.

stewart-ibm:
I'd tend to agree - compiler warnings++ (gcc6 is looking good... and we should probably start trying to build various things with gcc6 soon-ish).

I'll find out and report them to hostboot, hopefully with patches :)

deece:
The warning that aborts the build is new to GCC 5:
There are a number of instances of this unused itum in a header:
../../../src/include/usr/targeting/common/targetservice.H:137:22:
error: ‘TARGETING::MASTER_PROCESSOR_CHIP_TARGET_SENTINEL’ defined but
not used [-Werror=unused-variable]

shenki:
That warning is hostboot bug https://github.com/open-power/hostboot/issues/38

PR #347, opened 2016-02-10
Enable MTD support for Petitboot & bump version

sammj:
These patches add support for Petitboot to access MTD devices via libflash, and bump the Petitboot version to add the relevant support and several other minor fixes.


shenki:
LGTM, please merge

PR #359, opened 2016-02-24
Prepare for new buildroot

deece:
The updated buildroot infrastructure will complain about missing dependencies and not build skiboot (as it was not marked as a package).

Hostboot build will fail sporadically as the generated enums sometimes contain duplicate elements (perl bug?), sorting fixes (or at least hides) this problem.


deece:
submitted to the wrong branch

PR #360, opened 2016-02-24
Prepare for new buildroot

deece:
The updated buildroot infrastructure will complain about missing dependencies and not build skiboot (as it was not marked as a package).

Hostboot build will fail sporadically as the generated enums sometimes contain duplicate elements (perl bug?), sorting fixes (or at least hides) this problem.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/open-power/op-build/360)


williamspatrick:
Ok to test

Patrick

williamspatrick:
@deece - Can you create a pull request for the Hostboot repository with the fix you have here if you think it is valid?  We should not add patches to a project that are not also contributed to that project (and hopefully accepted).

@dcrowell77 - FYI.

deece:
Sure, I just need to reproduce the original problem. It’s taking some time as it’s an intermittent issue (duplicate enum elements being emitted), and I’m investigating in the background.

stewart-ibm:
So, the skiboot BR_TARGET_SKIBOOT to BR_PACKAGE_SKIBOOT fix does fix a long time bug with the openpower_mambo_defconfig configuration, where skiboot hasn't been being built. So I suspcet this is actually a good idea. It may be good to propose this change individually.

deece:
I think that was the issue I was investigating that triggered this change. I was reviewing the build log and noticed skiboot was not built, even though the build completed.

On 8 March 2016 11:43:38 AM AEDT, Stewart Smith <notifications@github.com> wrote:
>So, the skiboot BR_TARGET_SKIBOOT to BR_PACKAGE_SKIBOOT fix does fix a
>long time bug with the openpower_mambo_defconfig configuration, where
>skiboot hasn't been being built. So I suspcet this is actually a good
>idea. It may be good to propose this change individually.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/open-power/op-build/pull/360#issuecomment-193529698

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

stewart-ibm:
I think with the Hostboot issue being dealt with via the Hostboot tree, the BR2_TARGET -> BR2_PACKAGE change for skiboot already being in, and my soon-to-be-pull-requested rebase for buildroot 2016.02, this pull req should be closed?

@deece your thoughts?

deece:
Agreed

PR #361, opened 2016-02-24
Update openpower-pnor version

spinler:
This picks up the reprovisioning changes.


stewart-ibm:
I think we should instead:
1) pick up commit that merges these changes into open-power/pnor rather than the commit on the branch.
2) better describe what this change is. Something like:
"OpenPOWER PNOR layout XML now contains annotations as to what PNOR partitions should be erased versus kept when a machine is reprovisioned (i.e. "restore to factory default")"
3) Define somewhere what reprovision means, and how this is different to "restore to factory default settings" or something, and why this is different to the preserve tag.

williamspatrick:
Deprecated by #379.

PR #363, opened 2016-02-25
Set option to expand aliases

KennethWilke:
This `shopt` command allows the `op-build` alias to be expanded in non-interactive bash shells. I added this to my fork to allow easier build automation through Jenkins.

### DIFF
--- a/op-build-env
+++ b/op-build-env
@@ -8,4 +8,5 @@ fi
 export BR2_EXTERNAL=${__PWD}/openpower
 export BR2_DL_DIR=${__PWD}/dl
 
+shopt -s expand_aliases
 alias op-build="make --directory=${__PWD}/buildroot O=${__PWD}/output "


williamspatrick:
Ok to test.

williamspatrick:
Looks ok to me.  @stewart-ibm - any opinion?

stewart-ibm:
Having stumbled across this myself in my own jenkins playground, it's probably safe to do this here and ignore that nagging voice in my ear that there must be a good reason not to do it too.

Merged to master-next.

jazurin:
One of our nightly Jenkins jobs failed with the following:
+ shopt -s expand_aliases
/tmp/hudson9087134352605639138.sh: 11: ./op-build-ev: shopt: not found

It turns out that Jenkins defaults to sh when running shell scripts.   The simple fix was to add the Bash shebang line (#!/bin/bash).  Might be worth noting in the README.

williamspatrick:
Jay,

Can you add a pull request to force /bin/bash in that script?

On Fri, Feb 26, 2016 at 03:09:02PM -0800, Jay Azurin wrote:
> One of our nightly Jenkins jobs failed with the following:
> + shopt -s expand_aliases
> /tmp/hudson9087134352605639138.sh: 11: ./op-build-ev: shopt: not found
> 
> It turns out that Jenkins defaults to sh when running shell scripts.   The simple fix was to add the Bash shebang line (#!/bin/bash).  Might be worth noting in the README.
> 
> ---
> https://github.com/open-power/op-build/pull/363#issuecomment-189520870

-- 
Patrick Williams

KennethWilke:
I think the script he's referring to is the one you enter into the jenkins job interface. Possibly a better solution would be to set this thing up as a shell function instead of an alias, as it doesn't need to be expanded by an option and would work well in `/bin/sh` and `/bin/bash`. I can submit a pull request for that if that's the path you'd like to take.

PR #373, opened 2016-03-03
Shell function instead of alias for op-build

KennethWilke:
This is in response to the comments shared on PR #363 . It should skip the need for the `shopt` call and work for `/bin/sh` and non-interactive bash shells

@stewart-ibm @williamspatrick @jazurin how do you guys feel about this instead?


jazurin:
ok to test

stewart-ibm:
Looks good, merged to master-next.

PR #377, opened 2016-03-07
Add CI building scripts using Docker

stewart-ibm:
This is a first stab at a set of scripts to build each defconfig
on each supported build platform inside Docker containers.

The idea is that on x86-64 and ppc64le hosts, we can easily test
that everything still at least builds on our supported build
platforms.

Due to docker image availability and Docker's complete lack of
acknowledging that there may exist architectures other than x86-64, we
get to jump through some rather stupid hoops to work out what
to build where.

Tested on Fedora 23 x86-64 and Ubuntu xenial snapshot with docker
daily snapshot on ppc64le.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/open-power/op-build/377)


williamspatrick:
Your commit message says Ubuntu Xenial but the code says 14.04.  Where are the docker files for Xenial?

Why did we choose the latest Fedora but the LTS of Ubuntu?  Should we also build the latest Ubuntu?

stewart-ibm:
Xenial host, 14.04 in a container.

Looking to add CentOS 6 (as this is what power firmware builds on) and latest Ubuntu in future commit.

PR #382, opened 2016-03-09
Add additional packges to the VERSION partition

sammj:
This adds versions for the kernel and petitboot packages, and explicitly points out the op-build and buildroot versions.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/open-power/op-build/382)

### DIFF
--- a/openpower/package/VERSION.readme
+++ b/openpower/package/VERSION.readme
@@ -72,10 +72,12 @@
 
     Current sub packages included in openpower-pnor.mk $(OPENPOWER_VERSIONED_SUBPACKAGES)
 
-    - hostboot
-    - occ
     - skiboot
+    - hostboot
+    - petitboot kernel
+    - petitboot
     - $(platform)-xml
+    - occ
     - hostboot-binaries
     - capp-ucode
 
--- a/openpower/package/openpower-pnor/openpower-pnor.mk
+++ b/openpower/package/openpower-pnor/openpower-pnor.mk
@@ -33,7 +33,7 @@ OPENPOWER_PNOR_SCRATCH_DIR = $(STAGING_DIR)/openpower_pnor_scratch/
 OPENPOWER_VERSION_DIR = $(STAGING_DIR)/openpower_version
 
 # Subpackages we want to include in the version info (do not include openpower-pnor)
-OPENPOWER_VERSIONED_SUBPACKAGES = hostboot occ skiboot hostboot-binaries $(XML_PACKAGE) capp-ucode
+OPENPOWER_VERSIONED_SUBPACKAGES = skiboot hostboot linux petitboot $(XML_PACKAGE) occ hostboot-binaries capp-ucode
 OPENPOWER_PNOR = openpower-pnor
 
 define OPENPOWER_PNOR_INSTALL_IMAGES_CMDS
--- a/openpower/package/pkg-versions.mk
+++ b/openpower/package/pkg-versions.mk
@@ -71,7 +71,7 @@ echo -n "	$(1)-" > $$($(2)_VERSION_FILE)
 # If site local
 # Add site local and user, local commit, if local is dirty
 # Else not local
-# Add package version, op-build is dirty, op-build patches
+# Add package version, extraversion if linux, op-build is dirty, op-build patches
 if [ "$$($(2)_SITE_METHOD)" == "local" ]; then \
 echo -n "site_local-" >> $$($(2)_VERSION_FILE); \
 whoami | xargs echo -n >> $$($(2)_VERSION_FILE); \
@@ -88,6 +88,14 @@ else \
 [ `echo -n $$($(2)_VERSION) | wc -c` == "40" ] && (echo -n $$($(2)_VERSION) | \
 	sed "s/^\([0-9a-f]\{7\}\).*/\1/;s/$(1)-//;" >> $$($(2)_VERSION_FILE)) \
 	|| echo -n $$($(2)_VERSION) | sed -e 's/$(1)-//' >> $$($(2)_VERSION_FILE); \
+\
+if [ $(filter "LINUX", "$(2)") == "$(2)" ]; then \
+	if ls $$(BUILD_DIR)/$(1)-$$($(2)_VERSION)/Makefile 1>/dev/null; then \
+		head $$(BUILD_DIR)/$(1)-$$($(2)_VERSION)/Makefile | grep EXTRAVERSION \
+		| cut -d ' ' -f 3 | \
+		xargs echo -n >> $$($(2)_VERSION_FILE); \
+	fi; \
+fi; \
 \
 cd "$$(BR2_EXTERNAL)"; git describe --all --dirty | \
 	if grep -e "-dirty"; then \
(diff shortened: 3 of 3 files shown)


williamspatrick:
Reviewed 1 of 2 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[openpower/package/openpower-pnor/openpower-pnor.mk, line 36 \[r1\]](https://reviewable.io:443/reviews/open-power/op-build/382#-KCRZQJTiDP3Qxv4sbzc:-KCRZQJU-oU8Mf2TMcKz:221319476) ([raw file](https://github.com/open-power/op-build/blob/081ec56b0dd41a9e622eb5be1a3839bebf558a19/openpower/package/openpower-pnor/openpower-pnor.mk#L36)):*
Why this order?  Obviously there is some OPAL bias here by putting those packages all first. ;)

I would hesitate to generally change the order because this information is sent down to an IPMI sensor and people might be looking for some specific keywords in test.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/open-power/op-build/382)*

williamspatrick:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[openpower/package/pkg-versions.mk, line 177 \[r3\]](https://reviewable.io:443/reviews/open-power/op-build/382#-KCRab7SjCMw8csq90wF:-KCRab7SjCMw8csq90wG:-889215144) ([raw file](https://github.com/open-power/op-build/blob/41048d22aff1ebf39a163aeac6641777c992200e/openpower/package/pkg-versions.mk#L177)):*
Shouldn't we only do this if OPBUILD_VERSION was not empty?  In almost all cases this will be redundant.

---

*[openpower/package/pkg-versions.mk, line 185 \[r3\]](https://reviewable.io:443/reviews/open-power/op-build/382#-KCRb581caMIyBZqQPlW:-KCRb581caMIyBZqQPlX:-1643511960) ([raw file](https://github.com/open-power/op-build/blob/41048d22aff1ebf39a163aeac6641777c992200e/openpower/package/pkg-versions.mk#L185)):*
Why do we need the buildroot number if we have the op-build number?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/open-power/op-build/382)*

sammj:
> Why this order? Obviously there is some OPAL bias here by putting those packages all first. ;)

Haha, probably a little bit of bias. Mainly I took some suggestions on which versions are going to be more immediately useful if reading a bug report, which probably includes hostboot/skiboot/kernel versions especially. Suggestions from not-OPAL as to what should be prioritised very welcome.

> Shouldn't we only do this if OPBUILD_VERSION was not empty? In almost all cases this will be redundant.

Yes good point.

> Why do we need the buildroot number if we have the op-build number?

Ah right, yes that's probably superfluous then. @shenki it's not possible to have a build that doesn't include the buildroot subcommit specified by op-build right?

shenki:
On Thu, Mar 10, 2016 at 9:16 AM, Sam Mendoza-Jonas
<notifications@github.com> wrote:
>
> Ah right, yes that's probably superfluous then. @shenki it's not possible to have a build that doesn't include the buildroot subcommit specified by op-build right?

It is possible. If the person doing the build checks out their
op-build repository, and then manually checks out a copy of the
buildroot repository, then the op-build SHA gives us no information on
what the buildroot SHA is.

This is how the IBM op810.10 firmware is created.

williamspatrick:
On Wed, Mar 09, 2016 at 02:46:40PM -0800, Sam Mendoza-Jonas wrote:
> > Why this order? Obviously there is some OPAL bias here by putting those packages all first. ;)
> 
> Haha, probably a little bit of bias. Mainly I took some suggestions on which versions are going to be more immediately useful if reading a bug report, which probably includes hostboot/skiboot/kernel versions especially. Suggestions from not-OPAL as to what should be prioritised very welcome.
> 

My preference would be:
1. {skiboot hostboot linux} in any order.
2. buildroot
3. ${XML_PACKAGE} occ hostboot-binaries
4. capp-ucode

I realize it might be hard to put buildroot there so I would be ok with
it at either end.

-- 
Patrick Williams

stewart-ibm:
The tool producing what goes in the PNOR could just sort the list to be in alphabetical order. That's likely to make grumpy-debugging-us happier.

stewart-ibm:
(but I also think the bike shed should be blue :)

sammj:
> The tool producing what goes in the PNOR could just sort the list to be in alphabetical order. 

The problem as far as I've seen though is that the IPMI response will only return the first 6 or so versions, so we need to choose what goes up the top :)

williamspatrick:
On Wed, Mar 09, 2016 at 03:20:24PM -0800, Stewart Smith wrote:
> The tool producing what goes in the PNOR could just sort the list to be in alphabetical order. That's likely to make grumpy-debugging-us happier.

Except that the IPMI sensor that this data gets fed into is only so
long.  It will likely be cropped with the new content you are putting
in, thus getting the order "right" is somewhat important.

But, if you don't care about the IPMI sensor information because you're
going to get this information from the OS anyhow, please put all your
stuff at the end. ;)

> 
> ---
> https://github.com/open-power/op-build/pull/382#issuecomment-194561309

-- 
Patrick Williams

sammj:
Just pushed some changes - this reorders the output slightly (op-build/buildroot at top since mixing them is a bit trickier), and only adds op-build if it has been overridden.

Example:
```
Without override:
open-power-habanero-549b55e
        buildroot-2179a1c
        skiboot-5.1.13-opdirty
        hostboot-ce43144-5355e08
        linux-4.4.4
        habanero-xml-5a2a541-opdirty-0b04661
        occ-1093bf9-opdirty
        hostboot-binaries-7f593a3
        capp-ucode-105cb8f-opdirty

With override:
SOMEVENDOR-SOMEPLATFORM-v-1.2.3
        op-build-549b55e
        buildroot-2179a1c
        skiboot-5.1.13-opdirty
        hostboot-ce43144-5355e08
        linux-4.4.4
        habanero-xml-5a2a541-opdirty-0b04661
        occ-1093bf9-opdirty
        hostboot-binaries-7f593a3
        capp-ucode-105cb8f-opdirty
```

williamspatrick:
Do you know why the opdirty flag isn't consistent?

Patrick

sammj:
At @shenki 's suggestion an additional patch includes the linux extraversion in the string.

@williamspatrick I suspect this is because I've been rebuilding in the same tree, and the version scripts are re-using version files that already exist. I'll do a fresh build to be sure.

sammj:
A fresh build with a dirty op-build tree consistently adds the dirty tag (aside from buildroot)

```
open-power-habanero-ddc5fab-dirty
        buildroot-2179a1c
        skiboot-5.1.14-opdirty
        hostboot-ce43144-opdirty-5355e08
        linux-4.4.4-openpower1-opdirty-4b06ccb
        petitboot-72928ed-opdirty-91eed07
        habanero-xml-5a2a541-opdirty-0b04661
        occ-1093bf9-opdirty
        hostboot-binaries-7f593a3-opdirty
        capp-ucode-105cb8f-opdirty
```

shenki:
LGTM.

Can we change opdirty to EROR_FATAL_WARNING?

Or I_FAIL_AT_CONFIGURATION_MANAGEMENT?

williamspatrick:
Reviewed 1 of 1 files at r3, 3 of 3 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/open-power/op-build/382)*

PR #402, opened 2016-03-23
Update kernel defconfig - move to modules for near everything

stewart-ibm:
With CONFIG_KALLSYMS_ALL=n this is finally a win in total BOOTKERNEL size. This series reduces zImage.epapr by ~660kb

Tested on qemu and tuleta.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/402)


openpower/configs/linux/skiroot_defconfig
```diff
@@ -57,85 +57,89 @@ CONFIG_SYN_COOKIES=y
 # CONFIG_INET_XFRM_MODE_TUNNEL is not set
 # CONFIG_INET_XFRM_MODE_BEET is not set
 # CONFIG_IPV6 is not set
```
williamspatrick: We don't actually support IPv6?  Huh.  I'm surprised nobody has complained about that yet.  I thought it was a requirement in some circles these days.

stewart-ibm: Yeah, that should likely change Real Soon Now too, I was a bit surprised myself.

PR #407, opened 2016-03-25
Hostboot: Remove patch that restricted optimization.

williamspatrick:
At one point, Hostboot failed to run with -O3 when compiled with the
4.9.x series of compilers.  After internal investigation, we suspect
it was an issue with 4.9.0 specifically and 4.9.2+ appears to work
correctly.  We are using 4.9.3 here now.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/407)


williamspatrick:
This made it to runtime on both Habanero and Firestone, so it appears to be good.

PR #409, opened 2016-03-29
Updating Garrison XML pointer

Erich-Hauptli:
Picking up xml changes to tune Vdd gain and offset values per Jordan Keuseman.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/409)


williamspatrick:
@Erich-Hauptli - There is no need to submit these anymore.  We have garrision-xml on the list of packages that are automatically updated by a cronjob.  Already merged this package update in another commit.

PR #411, opened 2016-03-29
openpower/configs: Enable PETITBOOT_MTD by default

sammj:
Users expect to see platform versions in Petitboot now, so have support for it be enabled by default.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/411)


shenki:
LGTM

PR #412, opened 2016-03-30
openpower: Add loadkeys package to enable backtab

sammj:
This adds custom install rules for 'kbd' to install only the loadkeys utility. This allows us to add extra keymap definitions at init. Fixes https://github.com/open-power/petitboot/issues/8

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/412)


williamspatrick:
@shenki - Should we also submit this package upstream to buildroot?

shenki:
I had a chat with @sammj. He will look to upstream these.

This was an urgent request for barreleye, so I will leave it to Sam to make the call between shipping this as-is, or integrating an upstream patch.

PR #423, opened 2016-03-31
petitboot: depend on BR2_PACKAGE_LVM2 for devicemapper explicitly

stewart-ibm:
Previously, we added dependency to petitboot.mk but we were missing
it from Config.in, which buildroot 2016.02 errors out on.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/423)


stewart-ibm:
I wonder if the jenkins magic will pick up the changed commit.

Basically, I messed up the original patch and put in "depends" rather than "selects".

PR #425, opened 2016-03-31
Adding XZ decompression functionality for skiboot

e-liner:
We have already merged hostboot and open-power/pnor changes, this is the final change for RTC:125550.

This change will allow a user the option to XZ compress the skiboot image. The BR2_TARGET_SKIBOOT_XZ config will need to be enabled for the specific configuration, and then the skiboot image will be compressed. This change is only for compressing and decompressing the skiboot image section. If another section needs compression, we would have to enable decompression inside hostboot for that specific section, and make additional pnor changes.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/425)


e-liner:
(previous comments here: https://github.com/open-power/op-build/pull/418)

PR #430, opened 2016-04-06
Update garrison.mk

Erich-Hauptli:
Updated pointer to pick up Resonant clock frequencies such that they are in KHz instead of MHz.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/430)

### DIFF
--- a/openpower/package/garrison-xml/garrison.mk
+++ b/openpower/package/garrison-xml/garrison.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


jazurin:
@Erich-Hauptli garrison.mk is now pointing at dcf4876fafbadbdd9fc3b5a277ef8a13e1d72d6e
There is an automated process that updates the garrison.mk file with the latest change in garrison-xml, every night.  So you don't have to create these types of pull requests.

Thanks.

PR #433, opened 2016-04-07
Buildroot 2016.02

stewart-ibm:
Move to @shenki  buildroot 2016.2 tree as our buildroot.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/433)


shenki:
@stewart-ibm, any reason why you don't squash the two commits together?

Not that anyone actually does bisections of op-build :)

stewart-ibm:
> @stewart-ibm, any reason why you don't squash the two commits together?
>
> Not that anyone actually does bisections of op-build :)

Maybe one day we will? Plus, the first commit does work independently of
the second :)

-- 
Stewart Smith
OPAL Architect, IBM.

PR #441, opened 2016-04-14
Enable custom directory to work with hostboot and bootkernel changes

e-liner:
Previously, the hostboot directory wasn't pulling in changes from
the custom folder.  Also, we want to remove bootkernel and need to enable that by adding in a configuration flag check.

---

This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/441)


openpower/package/openpower-pnor/openpower-pnor.mk
```diff
@@ -61,6 +61,8 @@ define OPENPOWER_PNOR_INSTALL_IMAGES_CMDS
             -xz_compression $(BR2_OPENPOWER_PNOR_XZ_ENABLED)
 
         mkdir -p $(STAGING_DIR)/pnor/
+
+    if $(BR2_BOOTKERNEL_ENABLED); then \
```
williamspatrick: Rather than duplicating the whole blob, you should be able to do this by:

$(if $(findstring "true",$(BR2_BOOTKERNEL_ENABLED)),-bootkernel $(BINARIES_DIR)/$(BR2_LINUX_KERNEL))

openpower/package/hostboot/hostboot.mk
```diff
@@ -13,8 +13,15 @@ HOSTBOOT_DEPENDENCIES = host-binutils
 HOSTBOOT_INSTALL_IMAGES = YES
 HOSTBOOT_INSTALL_TARGET = NO
 
+# The BR2_HOSTBOOT_CONFIG_FILE variable has quotes around it, which throws off the wildcard function
+BR2_HOSTBOOT_CONFIG_FILE_NONQUOTED := $(shell echo $(BR2_HOSTBOOT_CONFIG_FILE))
```
williamspatrick: BR2_HOSTBOOT_CONFIG_FILE_NONQUOTED := $(subst ",,$(BR2_HOSTBOOT_CONFIG_FILE))

Using shell+echo here is dangerous.  Other than some unintended consequences if someone were to put &&, ||, or ; into the variable, it has the side effect of evaluating backslash escaped characters, which causes them to lose their backslash.

williamspatrick:
Two additional comments:
1. Please remove mention of the product name above.
2. Can you do a rebase to remove this "merge" commit?

e-liner:
Done and done. I knew there was a better way to do that openpower-pnor change. Thanks for the comments.

openpower/package/hostboot/hostboot.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-HOSTBOOT_VERSION ?= 787d61d8918730232ac546e0b679a19723fdd70a
+HOSTBOOT_VERSION ?= 32dff4ca9db6d8b088ed6ec63f147d2aaded3e69
```
williamspatrick: I don't think this change was intentional.  Seems to be rolling us back to an older Hostboot.

stewart-ibm:
@e-liner this has conflicts, able to resolve ?

stewart-ibm:
@e-liner is this still something you want merged? There's conflicts now, so things probably need to be fixed up.

stewart-ibm:
No activity for a long time, closing. Feel free to re-open if this is at some point in the future if this is needed.

PR #442, opened 2016-04-20
petitboot: Create default log directory

sammj:
With recent changes to Skiroot the petitboot-nc processes can start
before the pb-discover server. In this case /var/log/petitboot hasn't
been created by S15pb-discover yet, and each interface will fail to
fopen() its respective log file. This means each petitboot-nc process
will not produce logs the first time it is run.

Create the directory at build time to make sure we always get UI logs.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/442)


shenki:
LGTM

PR #446, opened 2016-04-25
Build PNOR refactor patches

scprek:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/446)

### DIFF
--- a/openpower/package/hostboot/hostboot-0003-fix-variable-syntax-in-fsp-makefile.patch
+++ b/openpower/package/hostboot/hostboot-0003-fix-variable-syntax-in-fsp-makefile.patch
@@ -1,0 +1,43 @@
+From 941e68e4dc5f8922bbe3a78e4f31a45a9a20fb74 Mon Sep 17 00:00:00 2001
+From: Stephen Cprek <smcprek@us.ibm.com>
+Date: Wed, 20 Apr 2016 17:25:35 -0500
+Subject: [PATCH] Fix TARGET_TEST syntax in makefile
+
+$TARGET_TEST always evaulated true
+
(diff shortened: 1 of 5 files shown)


stewart-ibm:
@williamspatrick I'll leave this for you considering all the hostboot touching.

I'd like to see a better description than just "refactoring". Are there any functional changes? How has it been checked that the refactoring ends up with the same result as before?

scprek:
So it's tricky because I have 3 commits internally that fail op-ci because they require a PNOR package change. So I need those commits to get reviewed internally (even though they fail op-ci), then modify this pull-request and make sure things all work. Then my hostboot commits can be merged and when op-build pulls them in, the patches will be removed. Perhaps cross dependencies can be handled better in the future for CI purposes. 

### Description: 
Currently much of the logic such as padding, ecc, headers etc are done in a variety of places 
(e.g. update_images.pl in PNOR, buildSbePart.pl in hostboot and others for FSP system use only). 
https://github.com/open-power/pnor/blob/master/update_image.pl#L121..L126

This requires changes in several places and unnecessary manual work when adding a new section and adding/removing XML tags.

This pull request is the start of the ongoing work to ultimately remove the update_image.pl script and have all PNOR processing (headers, padding, ecc) to be handled in buildpnor.pl. This way everything is data driven from the XML.

_**These changes are driven by Secureboot requirements.**_

stewart-ibm:
Yes, changing anything to do with PNOR in a non-trivial manner is a giant pile of mess. We're working on simplifying it and making all the code around it a lot more maintainable. Unfortunately, you probably need your changes in before we get everything to a better state.

scprek:
Going to do this differently to avoid co-reqs with openpower-pnor.

PR #452, opened 2016-04-28
linux: Disable CAPI support

sammj:
Support for the CAPI (CXL) driver in Skiroot is unnecessary and
potentially harmful as CAPI does not support being used over a kexec,
which could result in CAPI devices being unavailable in the host.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/452)


oohal:
Please do not merge this. I am planning on using the CAPI gzip accelerator card to enhance the boot wrapper decompresser.

mikey:
@oohal

williamspatrick:
Can I assume this was a joke?

Patrick

sammj:
My commit is legitimate; I think everyone else used this as an opportunity to test a few Github features :)

stewart-ibm:
Makes sense (well, not the damage we've done to @oohal due to working on the kernel wrapper).

PR #456, opened 2016-05-02
openpower/configs: Set default hostname to skiroot

sammj:
Change the default hostname to something more useful than "buildroot".

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/456)


mikey:
What about HostyMcHostFace?

shenki:
LGTM, please merge

ruscur:
A bit offtopic, but having some generic defconfig for every machine type would be good...changing something like this becomes more of a pain as you add more and more machines

shenki:
@ruscur that's super off topic. Please open a bug

PR #473, opened 2016-05-12
Merge master into master-next

stewart-ibm:
due to quirks of dev process, master is ahead of master-next by a few commits, making fixes on master while it's waiting for next release being... problematic.

This pull req merges in the old outstanding commits in master to master-next in one commit, and then the kernel update in another commit, creating a (slightly) clean history

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/473)


shenki:
Partially my fault. I sent a pull request some time ago for powerpc-utils that was against master. I did this because I thought that once master as tagged, we could use it as... master.

I propose we do that again from this cycle onwards, and use master-next as we did when it was first created - as a staging area for non-bugfix changes when we're about to tag a release on master.

Secondly, Patrick did ask me to send a PR for the new kernel to master and master-next. It slipped through the gaps, sorry.

PR #487, opened 2016-05-27
Bump Petitboot to v1.1.0

sammj:

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/487)

### DIFF
--- a/openpower/package/petitboot/petitboot.mk
+++ b/openpower/package/petitboot/petitboot.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-PETITBOOT_VERSION = v1.0.1
+PETITBOOT_VERSION = v1.1.0
 PETITBOOT_SITE ?= $(call github,open-power,petitboot,$(PETITBOOT_VERSION))
 PETITBOOT_DEPENDENCIES = ncurses udev host-bison host-flex lvm2
 PETITBOOT_LICENSE = GPLv2


williamspatrick:
retest this please

sammj:
@williamspatrick I suspect this failure is caused by a change in how the Petitboot UI starts, specifically: http://git.ozlabs.org/?p=petitboot;a=commit;h=bbaa3ac16e14b1f386d70d14f16d0164d163017e

Loooking through the Jenkins logs that appears to be roughly where it's failing, but I can't see the job configuration so I can't see what it is trying to do. Is there a way I can see that? Could be that the test will have to be updated.

stewart-ibm:
Sam Mendoza-Jonas <notifications@github.com> writes:
> @williamspatrick I suspect this failure is caused by a change in how the Petitboot UI starts, specifically: http://git.ozlabs.org/?p=petitboot;a=commit;h=bbaa3ac16e14b1f386d70d14f16d0164d163017e
>
> Loooking through the Jenkins logs that appears to be roughly where
> it's failing, but I can't see the job configuration so I can't see
> what it is trying to do. Is there a way I can see that? Could be that
> the test will have to be updated.

We should soon switch things over to run op-test-framework rather than
anything else, and fix anything there.


-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
@jazurin - Can you work with Sam on getting this to pass.

jazurin:
@sammj After a "S0/G0: working" state is read from the Host Status sensor,  the test simply pings the OS partition network interface every 30 seconds until it responds.  It tries that 6 times then it gives up.  The Petitboot UI seems to come up on the log, but nothing else after that.

sammj:
@jazurin Right, I see the failed pings in the console output now. There's not much to see beyond that though - hypothetically the network could not come up if pb-discover crashed but that doesn't seem to be the case from what I can see. Is it possible to rerun the test and interact on failure? Seeing the /var/log/petitboot/pb-discover.log file would be especially helpful.

sammj:
@jazurin Ping?

jazurin:
@sammj  sorry for the slow response.   I managed to save the resulting PNOR image and recreated the issue.  I was able to interact with petiboot and dump the pb-discover.log.  
An interesting snippet below and the whole file here: https://gist.github.com/jazurin/4845a5c6d255456c658b49377f71932b
```
Running command:
 exe:  /sbin/ip
 argv: '/sbin/ip' 'link' 'set' 'eth0' 'up'
ip: SIOCGIFFLAGS: No such device
failed to bring interface eth0 up
```

jazurin:
@sammj I left the system in the failing state.  You could start an sol session an interact with petitboot if you want to gather more logs.  I'll probably reclaim the system back tomorrow by noon so other tests can run.
ipmitool -H 9.3.29.140 -I lanplus -U ADMIN -P admin sol activate

sammj:
> failed to bring interface eth0 up

That may be a red herring, as udev will rename that device. Petitboot does appear to dhcp on the properly-named interface:

```
udhcpc (v1.24.2) started
 exe:  /sbin/udhcpc
 argv: '/sbin/udhcpc' '-R' '-f' '-O' 'pxeconffile' '-O' 'pxepathprefix' '-p' '/var/petitboot//udhcpc-enP2p1s0f2.pid' '-i' 'enP2p1s0f2' '-x' '0x5d:000e'
```

However dhcp appears to fail regardless, eg:
```
# udhcpc -i enP2p1s0f2 --now -f 
udhcpc (v1.24.2) started
Sending discover...
Sending discover...
Sending discover...
No lease, failing
```

More to the point, none of the changes in Petitboot are related to network. Is there a problem with the network? Is it relevant that this only failed on the Habanero, and not the other platforms?

williamspatrick:
The Austin lab doesn't do DHCP. Is there an nvram or BMC setting that is now being ignored for static configs?

sammj:
From the pb-discover.log file above it appears there is no static network config saved - @jazurin could this have been cleared?

jazurin:
@sammj   We had many CI runs after this one and have not seen problems with this habanero booting all the way to the OS.  
Also about the other systems passing.  At moment only our habanero is configured to run the test that boots all the way to the OS, the other systems only boot to petitboot.

sammj:
> @sammj We had many CI runs after this one and have not seen problems with this habanero booting all the way to the OS.

Oh so it's waiting for the OS to respond not Petitboot? i.e Petitboot not having network isn't the problem in this case.

> Also about the other systems passing. At moment only our habanero is configured to run the test that boots all the way to the OS, the other systems only boot to petitboot.

@jazurin When I was logged on to this machine I noticed an IPMI Network override was present - this would have stopped it from booting an OS from disk. Was that intentional? Is that machine still available to debug?

jazurin:
@sammj We didn't apply any IPMI network override intentionally.  
Is there a dependency on certain AMI FW level?
We haven't updated the BMC FW on this system in a while:
gfw172$ ssh -k sysadmin@9.3.29.140 "cat /proc/ractrends/Helper/FwInfo"
sysadmin@9.3.29.140's password:
FW_VERSION=2.00.80238
FW_DATE=Jul  6 2015
FW_BUILDTIME=13:35:48 EDT
FW_DESC=HABANERO P2 SRC RR9 07062015
FW_PRODUCTID=1
FW_RELEASEID=RR9
FW_CODEBASEVERSION=2.X
gfw172$

You can use the system tonight to debug if needed.

jazurin:
9.3.29.140 is left in failed state for debug. Will reclaim tomorrow at noon.

sammj:
> 9.3.29.140 is left in failed state for debug. Will reclaim tomorrow at noon.

Thanks @jazurin 

Looking more closely, I think the machine is hitting an existing bug in Petitboot to do with IPMI overrides - it's not introduced by this but it seems the planets aligned just right to break the test :)

I'll include the fix in a new point release and make a new PR for that instead - thanks for helping debug.

Issue #507, opened 2016-06-27
op-build master-next broken on rhel6.4 pool systems , missing 'makeinfo'


jazurin:
Our op-build CI is failing on rhel6.4 nodes.    I believe it started happening after https://github.com/open-power/op-build/pull/504 was merged.  The reason the pull request CI didn't catch this is because I had turned off testing on rhel6.4 machines temporarily due to the github download bug affecting only our rhel6.4 nodes.

```
>>> host-binutils 2.26 Building
[...]

PR #510, opened 2016-06-29
petitboot: Bump version to 1.2.0

sammj:

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/510)

### DIFF
--- a/openpower/package/petitboot/petitboot.mk
+++ b/openpower/package/petitboot/petitboot.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-PETITBOOT_VERSION = v1.1.1
+PETITBOOT_VERSION = v1.2.0
 PETITBOOT_SITE ?= $(call github,open-power,petitboot,$(PETITBOOT_VERSION))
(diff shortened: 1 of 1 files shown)


sammj:
Ah, looks like the test is running into the same issue as https://github.com/open-power/op-build/issues/507

```
/tmp/hostboot/jenkins-slave/workspace/op_build_ghprb_trigger/output/build/host-binutils-2.26/missing: line 81: makeinfo: command not found
WARNING: 'makeinfo' is missing on your system.
         You should only need it if you modified a '.texi' file, or
         any other file indirectly affecting the aspect of the manual.
         You might want to install the Texinfo package:
         <http://www.gnu.org/software/texinfo/>
         The spurious makeinfo call might also be the consequence of
         using a buggy 'make' (AIX, DU, IRIX), in which case you might
         want to install GNU make:
         <http://www.gnu.org/software/make/>
```

williamspatrick:
@sammj - Please rebase and this will likely pass CI now.

PR #511, opened 2016-06-29
Update OCC build name to op_occ_160628a

wilbryan:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/511)


williamspatrick:
You don't need to make these kind of pull requests.  We have a cronjob that does it for OCC automatically.

williamspatrick:
See #509 as an example.

PR #528, opened 2016-07-13
Source env.bash when running hbDistribute

scprek:
hbDistrubte will determine if secureboot functionality was compiled
by looking in obj/genfiles/config.h. To simplify getting to that file
we can source env.bash once in the hostboot directory

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/528)

### DIFF
--- a/openpower/package/hostboot/hostboot.mk
+++ b/openpower/package/hostboot/hostboot.mk
@@ -23,7 +23,7 @@ define HOSTBOOT_BUILD_CMDS
 endef
 
 define HOSTBOOT_INSTALL_IMAGES_CMDS
-        cd $(@D) && $(@D)/src/build/tools/hbDistribute --openpower $(STAGING_DIR)/hostboot_build_images/
+        cd $(@D) && source ./env.bash && $(@D)/src/build/tools/hbDistribute --openpower $(STAGING_DIR)/hostboot_build_images/
 endef
 
 $(eval $(generic-package))


williamspatrick:
Does make guarantee to create a new $(SHELL) session for each recipe?  Do we need to wrap the whole call in a 'bash -e' now to make sure we are not polluting an environment outside of that one line?

scprek:
So I just need a way for hbDistribute to get the hostboot top-level-dir. Currently it has `${PROJECT_ROOT}` which is set via env.bash. I am positive there's be a way to not source , but I was struggling and saw that build simply sourced env.bash to solve those problems.
If I had a way to get the top level of hostboot then in hbDistrubute I'd need a 
` if [ ${OPENPOWER_BUILD}=1 ]` check and use an op-build env variable.

Since the hostboot directory in output is actually not a .git repo anymore, the git top level command gives me op-build.

williamspatrick:
On Wed, Jul 13, 2016 at 09:19:58AM -0700, Stephen Cprek wrote:

> If I had a way to get the top level of hostboot then in hbDistrubute I'd need a 
> ` if [ ${OPENPOWER_BUILD}=1 ]` check and use an op-build env variable.

I would not recommend making that kind of change in the Hostboot
repository.  We should try to avoid polluting package repositories with
"I know I'm being built under this special build system so do XYZ
instead."  It isn't maintainable if we ever decide to build the firmware
with a build system other than Buildroot.

-- 
Patrick Williams

jazurin:
retest this please

williamspatrick:
@scprek - Test it by calling 'env' in the next line and see if it has any pollution from the previous line's "source" call.

scprek:
Tested with the following andthe env variables do not escape the call.

```makefile
define HOSTBOOT_INSTALL_IMAGES_CMDS
        cd $(@D) && source ./env.bash && $(@D)/src/build/tools/hbDistribute --openpower $(STAGING_DIR)/hostboot_build_images/
        echo "steve2"
        env
endef
```

PR #531, opened 2016-07-18
linux: use the opal console inside zImage wrapper

oohal:
The zImage wrapper currently does not produce any output on OPAL
machines. This patch adds an OPAL console backend to the wrapper
so that errors (such as decompression errors) can be reported to
users.

Signed-off-by: Oliver O'Halloran <oohall@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/531)


jazurin:
ok to test

williamspatrick:
@shenki - Need your approval on this.

shenki:
The PR is for master. Is that intentional?

oohal:
Master was intentional since this is more of a bugfix than a feature. I'm not particularly fussed if you want to put it in next though.

williamspatrick:
Ah, yeah, we need this PR against master-next.  We only put stop-ship / critical security level fixes into master.  Everything else goes into master-next.  We expect firmware and system vendors to be doing their own stable streaming of 1.x.y releases and we tag 1.x releases.

shenki:
@oohal I'll pull this into the kernel tree. Leave this open until I've done that

shenki:
Resolved as of https://github.com/open-power/op-build/commit/907068c54c76d655486175e2e06484e69d1d1485

PR #532, opened 2016-07-18
Export PETITBOOT_VERSION to Petitboot autoreconf step

sammj:
This allows us to get the full sha-dirty-patch version string in Petitboot.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/532)

### DIFF
--- a/openpower/package/petitboot/petitboot.mk
+++ b/openpower/package/petitboot/petitboot.mk
@@ -19,6 +19,8 @@ PETITBOOT_CONF_OPTS += --with-ncurses --without-twin-x11 --without-twin-fbdev \
 	      HOST_PROG_SHUTDOWN=/usr/libexec/petitboot/bb-kexec-reboot \
 	      $(if $(BR2_PACKAGE_BUSYBOX),--with-tftp=busybox)
 
+PETITBOOT_AUTORECONF_ENV += PETITBOOT_VERSION=`cat $(PETITBOOT_VERSION_FILE) | cut -d '-' -f 2-`
+
 ifdef PETITBOOT_DEBUG
 PETITBOOT_CONF_OPTS += --enable-debug
 endif
--- a/openpower/package/pkg-versions.mk
+++ b/openpower/package/pkg-versions.mk
@@ -114,7 +114,7 @@ endef ###
 
 # Add appropriate templates to hooks
 $(2)_POST_PATCH_HOOKS += $(2)_OPENPOWER_PATCH_FILE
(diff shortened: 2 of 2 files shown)


openpower/package/petitboot/petitboot.mk
```diff
@@ -19,6 +19,8 @@ PETITBOOT_CONF_OPTS += --with-ncurses --without-twin-x11 --without-twin-fbdev \
 	      HOST_PROG_SHUTDOWN=/usr/libexec/petitboot/bb-kexec-reboot \
 	      $(if $(BR2_PACKAGE_BUSYBOX),--with-tftp=busybox)
 
+PETITBOOT_AUTORECONF_ENV += PETITBOOT_VERSION=`cat $(PETITBOOT_VERSION_FILE) | cut -d '-' -f 2-`
```
williamspatrick: I must be missing why do you want to use a different version variable than ${PETITBOOT_VERSION} from above?

Seems like this could become a regularly repeated pattern.  Do we need to do it for other packages?

williamspatrick:
@scprek  - Please review.

openpower/package/petitboot/petitboot.mk
sammj: The buildroot variable ${PETITBOOT_VERSION} is just "v1.2.1" for example, whereas the version file includes the full "v1.2.1-(opdirty?)-(patchversion)" string.
Skiboot at least does something similar here: https://github.com/open-power/op-build/blob/master/openpower/package/skiboot/skiboot.mk#L30

scprek: I'm still  confused how `PETITBOOT_AUTORECONF_ENV` is needed and it's potential use?

the skiboot reference and others like https://github.com/open-power/op-build/blob/master/openpower/package/hostboot/hostboot.mk#L19 just overwrite the same variable.

sammj: Skiboot (and it looks like Hostboot too?) sets an environment variable (eg. SKIBOOT_VERSION) to be recognised at build time. Petitboot however determines it's version string during the autoreconf step before configuring (ie. in AC_INIT), so we need to set this variable such that it is recognised *then* rather than at build time.

sammj:
Any other thoughts on this?

stewart-ibm:
This looks good to me, tested with a clean build and got  "Petitboot (v1.2.1-d695626)                      IBM PowerNV (emulated by qemu)" in my qemu test. So, looks good to merge!

PR #548, opened 2016-07-22
openpower/package: DEPENDENCIES is a list of packages, not targets

### LINKED ISSUE #547: legal-info target is broken
The legal-info target doesn't build (and hasn't for some time):

```
$ op-build legal-info
[verbose output omitted]
make[1]: *** No rule to make target 'common-p8-xml-install-images-all-legal-info', needed by 'firestone-xml-all-legal-info'.  Stop.
Makefile:36: recipe for target '_all' failed
make: *** [_all] Error 2
make: Leaving directory '/home/jk/IBM/openpower/op-build/buildroot'
```

Fix coming.

jk-ozlabs:
The current $target-xml packages use target names in their dependencies
lists, where they should be using package names. This breaks certain
targets, like 'legal-info', which create dependencies by adding other
suffixes to the package names.

From the  buildroot manual:

  These dependencies are listed in terms of lower-case package names,
  which can be packages for the target (without the host- prefix) or
  packages for the host (with the host-) prefix). Buildroot will ensure
  that all these packages are built and installed before the current
  package starts its configuration.

This change fixed these DEPENDENCIES defintions to use package names
instead. Consequently, we don't need to filter out the -install-images
target suffix in the op-target-dependencies script.

Fixes #547.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/548)

### DIFF
--- a/openpower/package/barreleye-xml/barreleye-xml.mk
+++ b/openpower/package/barreleye-xml/barreleye-xml.mk
@@ -8,7 +8,7 @@ BARRELEYE_XML_VERSION ?= 81ac3ff3c4ccd16d2174f455c1cb17976daca948
 BARRELEYE_XML_SITE = $(call github,open-power,barreleye-xml,$(BARRELEYE_XML_VERSION))
 
 BARRELEYE_XML_LICENSE = Apache-2.0
-BARRELEYE_XML_DEPENDENCIES = hostboot-install-images openpower-mrw-install-images common-p8-xml-install-images
+BARRELEYE_XML_DEPENDENCIES = hostboot openpower-mrw common-p8-xml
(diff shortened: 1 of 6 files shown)


stewart-ibm:
Hrm... it appears that as a side effect we're missing a dependency around openpower-pnor needing zImage.papr already built... although there's an earlier error:

 Makefile:480: *** common-p8-xml is in the dependency chain of habanero-xml that has added it to its _DEPENDENCIES variable without selecting it or depending on it from Config.in.  Stop.

jk-ozlabs:
Yep, looks like there are missing dependencies here. I'll take a look.

PR #550, opened 2016-07-25
RFC: Fix dependencies for openpower-pnor & xml packages

jk-ozlabs:
Currently, there are a few issues with the dependency & configuration definitions for the PNOR and XML packages. This is preventing us from using some of the non-build targets, like `graph-depends` and `legal-info`, which don't currently work.

This proposed change populates the proper dependency info.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/550)

### DIFF
--- a/openpower/package/Config.in
+++ b/openpower/package/Config.in
@@ -1,6 +1,8 @@
 source "$BR2_EXTERNAL/package/openpower-ffs/Config.in"
 source "$BR2_EXTERNAL/package/hostboot/Config.in"
 source "$BR2_EXTERNAL/package/hostboot-binaries/Config.in"
+source "$BR2_EXTERNAL/package/openpower-mrw/Config.in"
+source "$BR2_EXTERNAL/package/common-p8-xml/Config.in"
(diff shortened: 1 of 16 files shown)


jk-ozlabs:
This seems to work fine with the new `firenze_defconfig` too:

```
[jk@pecola op-build]$ ls -l output/images/
total 118924
-rw-r--r-- 1 jk jk 52980736 Jul 25 13:52 rootfs.cpio
-rw-r--r-- 1 jk jk 10668664 Jul 25 13:52 rootfs.cpio.xz
-rw-r--r-- 1 jk jk 53452800 Jul 25 13:52 rootfs.tar
-rwxr-xr-x 1 jk jk   924854 Jul 25 13:52 skiboot.lid
-rw-r--r-- 1 jk jk  3848536 Jul 25 13:49 zImage.epapr
```

stewart-ibm:
This looks good to me, after discussing with @jk-ozlabs , we figure this is good to go.

Verified that the non build targets now work correctly.

Issue #538, opened 2016-07-19
mlx5_core driver doesn't survive kexec, IB cards not accessable


stewart-ibm:
Without suitable kernel patch, IB card doesn't work after kexec.

Solution is we either bring in a patch that has been discussed internally to IBM into our kernel tree, or we disable mlx5_core for v1.10

shenki:
I propose we add the patch that was sent to the mailing list:

 https://lists.ozlabs.org/pipermail/openpower-firmware/2016-July/000047.html

It's a backport of the upstream patch that adds kexec. Carol has tested it and it resolves the issue.

I have added the patch to my queue, but a release is blocked on getting the usb situation sorted for v4.4.15.

If we want to rush this through, I could release v4.4.14-openpower2 with the queued up fixes.

stewart-ibm:
Since this has gone into 4.4.15-openpower1 which is now sitting in master-next, closing this bug.

stewart-ibm:
So, it turns out it doesn't work, so as discussed on phone with @shenki and @sammj  - we're going to disable it for this release.

CC: @DEADBEEFh

sammj:
For reference, this is what I saw with the 4.4.15-openpower1 kernel on a machine with a Mellanox CX-4 card.

```
The system is going down NOW!t
Sent SIGTERM to all processes
Sent SIGKILL to all processes
cpu 0x0: Vector: 300 (Data Access) at [c000001fd73436a0]
    pc: d0000000047f6ff8: reclaim_pages+0x164/0x198 [mlx5_core]
    lr: d0000000047f7004: reclaim_pages+0x170/0x198 [mlx5_core]
    sp: c000001fd7343920
   msr: 9000000000009033
   dar: c000002000000000
 dsisr: 40000000
  current = 0xc000001fd728d900
  paca    = 0xc00000000fe80000   softe: 0        irq_happened: 0x01
    pid   = 2336, comm = kexec
enter ? for help
[c000001fd73439c0] d0000000047f79e8 mlx5_reclaim_startup_pages+0x8c/0x144 [mlx5_core]
[c000001fd7343a70] d0000000047f0bf8 mlx5_unload_one+0x1cc/0x234 [mlx5_core]
[c000001fd7343b10] d0000000047f0d38 mlx5_pci_err_detected+0xd8/0x180 [mlx5_core]
[c000001fd7343b80] c000000000209cac pci_device_shutdown+0x50/0xa0
[...]

PR #479, opened 2016-05-18
Add mlx5_core with ethernet support to petitboot to support 25/100 GbE devices

DEADBEEFh:
A few changes to the skiroot_defconfig.
Initial build and test looks good on firestone system.
mlx5_core.ko = 246088 bytes

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/479)

PR #558, opened 2016-08-01
disable mlx5 driver in skiroot_defconfig

stewart-ibm:
Quite simply, the driver doesn't work properly. It's a kexec disaster
and without (yet another) patch we have to carry locally, it's not
going to work for the imminent op-build v1.10.

We'll think about re-enabling this once the appropriate patches make
it to the Linux stable tree.

Fixes: https://github.com/open-power/op-build/issues/538
Reverts: https://github.com/open-power/op-build/pull/479
Reverts: 17b4c3c6e6389c8d5d354ef778ce1f39067f8a34

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/558)


shenki:
+1

PR #566, opened 2016-08-03
Update OCC FW 8-3-2016

wilbryan:
This update provides additional FFDC debug for OCC doorbell.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/566)

### DIFF
--- a/openpower/package/occ/occ.mk
+++ b/openpower/package/occ/occ.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


williamspatrick:
@wilbryan - Did you intend to push this to master?  Generally we don't need to do OCC updates because we have a job that does it automatically.

wilbryan:
Ah, so all I need to do is push it to the occ repository and the job will sync it with master/master-next?

williamspatrick:
It will sync with master-next. Master is only for critical fixes on top of tagged releases.

PR #568, opened 2016-08-05
Update capp-ucode to latest master

mikey:
Based on (IBM internal bug tracker) SW322717
"SUE may be presented to OS in rare situations during a CAPI workload if EEH encountered"

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/568)


stewart-ibm:
Marking for stable tree.

I haven't yet written up exactly how that's going to work, but in order to get into the IBM release I think somebody wants this in, it has to go there after it's hit master-next.

PR #579, opened 2016-08-11
linux: Release v4.4.17-openpower1

shenki:
Upstream 4.4.17 stable release.

Aslo adds a patch from Mamatha for changing the VGA ownership register
value to fix slow BMC SOL. Note that it is unclear as to why this patch
fixes the issue. However, it has been requested that we include it so
here it is.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/579)


stewart-ibm:
Marking as candidate for stable, depending on how it goes. I'd really like more justification about why this magic patch fixes things before pulling into stable (and we should consider our options over the next couple of weeks before op-build v1.11)

PR #582, opened 2016-08-15
Disabling SPD Writes for Garrison

aalugore:
-unset DJVPD_WRITE_TO_HW in garrison.config

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/582)


williamspatrick:
Why?

dcrowell77:
The RAS and system owners have determined that we should not be saving bit repair information permanently into the DIMM's SPD based on feedback from other OpenPOWER systems.  Instead we will preserve the repairs inside the VPD cache only.

williamspatrick:
Andy, please update the commit message to this effect.

aalugore:
Done.

williamspatrick:
Please clean up the commit history.  We don't need 3 commits for a 1 line change. ;)

aalugore:
Done. It seems there is still an ongoing discussion on whether or not everyone wants this by default. Please hold off on merge until further notice.

aalugore:
This can be merged (once we figure out OP CI HW)

From Michael Mueller:

Just got off the phone with Sertac.
We agree that the "Disable DIMM VPD udate" function is desired for Garrison too.
Our recommendation is to put the fix in, if it is easy and simple to put into the code now, without breaking other function and without a bunch of follow on defects because it might not work as expected.
Otherwise target for next convenient fix release.

stewart-ibm:
retest this please

jazurin:
@dcrowell77 why is this a garrison only change?
https://github.com/open-power/op-build/blob/master-next/openpower/configs/hostboot/habanero.config
https://github.com/open-power/op-build/blob/master-next/openpower/configs/hostboot/firestone.config
have:
 ```set DJVPD_WRITE_TO_HW```
 should those be changed too?

dcrowell77:
There is no technical reason, purely a decision based on release schedules and support/test plans.  Habanero and Firestone are already out in the wild with the current setting, whereas Garrison has time to go through additional testing after the change.

I don't know about barreleye schedules though, that might be worth updating.  @williamspatrick Do you have a good contact for Barreleye?

williamspatrick:
@johnhihi is the Barreleye contact.

PR #583, opened 2016-08-16
Add quiet to skiroot kernel command line

stewart-ibm:
Some BMCs have horrifically slow vuarts (or have enough debug options enabled
so that the vuart is horrifically slow), so slow that a 56k modem would
be an upgrade. Due to the amount of early boot messages that Linux outputs,
this can slow down boot a *lot* - in the order of *over* 20 seconds.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/583)


sammj:
Once they've been merged we should switch to using the skiroot/nvram cmdline update by @oohal (http://patchwork.ozlabs.org/patch/656528/ et al)

PR #589, opened 2016-08-19
Adding in new attribute types files

e-liner:
Hostboot is introducing a new attribute file for attributes that are openpower only, and the FSP side is unaware of.  On the hostboot side, these files are only compiled with a configuration flag: CONFIG_OPENPOWER_ONLY.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/589)


williamspatrick:
Retest this please.

Issue #500, opened 2016-06-21
repository per platform for MRW XML doesn't scale


stewart-ibm:
If we look into the future with 50-100 platforms, that many git repositories may be getting a bit hard to handle, and we then have a _lot_ of packages in our buildroot overlay to maintain, all of which will be roughly identical.

One possibility would be to have a single MRW repository and merge everything into it.

This would also make the process of upstreaming a new platform a lot simpler.

williamspatrick:
The more important step in my mind is to reduce the duplication per machine. 

We have:
1. The xml package itself. 
2. Buildroot defconfig. 
3. Hostboot config. 

The XML package can be simplified with a common .mk file to reduce that duplication. 

The defconfig we have the work from Sam that is awaiting upstream acceptance. 

The Hostboot config we should have someone from that team take on. 

I don't honestly see separate repositories as being a deal breaker, but I do see these duplications as painful. 

We are planning to use the XML to feed into openbmc as well, but we will obviously need more data. I'm not sure what the appropriate repository breakdown would be with that in mind. 

There is also no reason why XML repositories need to be in the open-power project. I would be fine pointing to company owned / maintained repositories elsewhere.

stewart-ibm:
I don't disagree with any of the above. Perhaps a change to our buildroot overlay so that it's just PLATFORM_XML rather than (HABANERO|FIRESTONE|PALMETTO)_XML would de-duplicate a lot.

stewart-ibm:
Considering that we've gone to one set of mk files for the XML, I'll close this issue for now.

PR #602, opened 2016-08-26
Add witherspoon-xml support

ploetzma:
Initial makefile for witherspoon-xml

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/602)


stewart-ibm:
We're getting a *lot* of these XML repositories and packages now. I think it's about time to consolidate and stop the duplication.

dcrowell77:
What are you proposing @stewart-ibm ?  A single project with a bunch of system files?  That wouldn't work from a permission/management standpoint.

williamspatrick:
The 'machine.mk' file is effectively a copy / paste between all the machines. We can certainly have a include.mk that reduces all that duplication.

All of the config variables could be consolidated too so they don't need to be uniquely named per machine.

stewart-ibm:
> What are you proposing @stewart-ibm ?  A single project with a bunch
> of system files?  That wouldn't work from a permission/management
> standpoint.

A single repository with a maintainer could certainly work, system
owners could just submit pull requests to it - and that process may
encourage people to notice other issues, including things that could be
simplified.

With 100 platforms, 100 different repositories will be beyond pretty
unweildly.

-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
I don't see how 1 repository with a maintainer vs 100 repositories owned by the system engineer is "unweildly."  The more important thing is for us to stop making backwards incompatible changes to the MRW such that we have to change all the repos "at once".  The issue I see with having a single "maintainer" is that there is no person who knows all of the attributes and all of the machines and all of the desired behavior by all the ODM/OEMs of the various machines.  Having one person maintain the single machine seems to be the closest thing we have to an expert that can satisfy these needs.

dcrowell77:
> stop making backwards incompatible changes to the MRW such that we have to change all the repos "at once"
It isn't a perfect fix, but I've heard some plans to add some explicit messages to the serverwiz2 gui that forces users to acknowledge and deal with changes when new attributes are added.

> With 100 platforms, 100 different repositories
This could change, but based on the current history I don't see most OP systems having any github presence at all.  Most ODM/OEMs are keeping their system files internally (for a variety of perfectly valid reasons).  That means the only system repos we would have would be for reference platforms provided by IBM or other partners.

williamspatrick:
Issue #500 is already open to discuss refactoring the MRW packages.  It is not the responsibility of @ploetzma to implement that feature just because he is inserting the "next" system.

stewart-ibm:
> I don't see how 1 repository with a maintainer vs 100 repositories
> owned by the system engineer is "unweildly."  The more important thing
> is for us to stop making backwards incompatible changes to the MRW
> such that we have to change all the repos "at once".  The issue I see
> with having a single "maintainer" is that there is no person who knows
> all of the attributes and all of the machines and all of the desired
> behavior by all the ODM/OEMs of the various machines.  Having one
> person maintain the single machine seems to be the closest thing we
> have to an expert that can satisfy these needs.

Agreed, incompatible changes to MRW aren't good.

I'm not suggesting that that maintainer does all the work, they may just
be able to co-ordinate it.

Plus, it would reduce the patch size to add a new machine, you could
conceivably then do it just with a new defconfig.

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
>> With 100 platforms, 100 different repositories
> This could change, but based on the current history I don't see most
> OP systems having any github presence at all.  Most ODM/OEMs are
> keeping their system files internally (for a variety of perfectly
> valid reasons).  That means the only system repos we would have would
> be for reference platforms provided by IBM or other partners.

This should massively change for Secure Boot though - it's not exactly
secure if you can't work out if you've been provided with backdoored
firmware :)

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
> Issue #500 is already open to discuss refactoring the MRW packages.
> It is not the responsibility of @ploetzma to implement that feature
> just because he is inserting the "next" system.

That's fair.

I'd give @ploetzma extra bonus points for doing it though :)

-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
```This should massively change for Secure Boot though - it's not exactly
secure if you can't work out if you've been provided with backdoored
firmware :)```

Secureboot really has nothing to do with "security" in the "I know exactly what is in this code and there are no security holes" sense.  Secureboot ensures "I am running code I expect to be running."  Vendor gives you an image with a signature that your system trusts and it only runs that.  If you have a TPM, you can also confirm the signatures from the TPM.

If a vendor puts a backdoor into an image you need to simply not trust their key anymore.  Of course, if they only have a vendor key on your system there isn't much you can do about it.

None of that _requires_ open source though.

PR #626, opened 2016-09-07
openpower/configs: Fix firenze_defconfig

sammj:
Commit e5d3bb2 "buildroot: Move to 2016.08 release" removed several
options from firenze_defconfig that weren't actually redundant. This
patch updates firenze_defconfig to an actual defconfig generated by
savedefconfig which disables unneeded options correctly.
This also disables p8-pore-binutils which isn't required if not building
the occ package.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/626)


sammj:
As a side note, it would be a good idea to add the firenze platform to the automated checks (even if just a build test) to catch this in the future.

shenki:
LGTM.

@williamspatrick can we please have this target (and all of the new ones?) added to the CI job that is run on op-build PRs?

stewart-ibm:
It may be @geissonator who needs to add it.

Maybe the auto-foo should use the ci/ scripts that even test build dependencies and building in supported OSs through the magic of docker.

williamspatrick:
The system that does the CI testing right now is incapable of running Docker. I would like to see both it and the openbmc test runs be based on a config file within the repo of "officially supported machine configs". 

@jazurin takes care of the OpenPower CI.

jazurin:
@shenki I will add firenze, openpower_mambo, openpower_p9_mambo to the PR build script, and witherspoon eventually when we can build it.  Thanks.

shenki:
Thanks @jazurin 

I didn't realise that witherspoon failed to build. Hopefully in the future we do not add platforms that do not build.

stewart-ibm:
@jazurin FWIW in my testing I just build every openpower/configes/*_defconfig

williamspatrick:
In theory that is what we would do, but the way the test job is organized in Jenkins, there is a sub-job per machine. That way they run in parallel. 

There is probably a way with a paramaterized build to do a wildcard but it isn't currently done like that.

stewart-ibm:
I remember using wildcard parameterized builds in Jenkins once... I quickly changed away from doing it :)

PR #629, opened 2016-09-09
Adding support for new attribute files.

e-liner:
These changes add support for the new optional hostboot attribute
files that are only available on openpower machines.

---

This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/629)

### DIFF
--- a/openpower/package/barreleye-xml/barreleye-xml.mk
+++ b/openpower/package/barreleye-xml/barreleye-xml.mk
@@ -40,9 +40,11 @@ define BARRELEYE_XML_BUILD_CMDS
             $(MRW_HB_TOOLS)/attribute_types.xml \
             $(MRW_HB_TOOLS)/attribute_types_hb.xml \
             $(MRW_HB_TOOLS)/attribute_types_openpower.xml \
+            $(MRW_HB_TOOLS)/attribute_types_oppowervm.xml \
             $(MRW_HB_TOOLS)/target_types_merged.xml \
             $(MRW_HB_TOOLS)/target_types_hb.xml \
             $(MRW_HB_TOOLS)/target_types_openpower.xml \
+            $(MRW_HB_TOOLS)/target_types_oppowervm.xml
             $(MRW_SCRATCH)/$(BR2_BARRELEYE_MRW_XML_FILENAME) > $(MRW_HB_TOOLS)/temporary_hb.hb.xml;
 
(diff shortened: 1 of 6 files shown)


openpower/package/barreleye-xml/barreleye-xml.mk
```diff
@@ -40,9 +40,11 @@ define BARRELEYE_XML_BUILD_CMDS
             $(MRW_HB_TOOLS)/attribute_types.xml \
             $(MRW_HB_TOOLS)/attribute_types_hb.xml \
             $(MRW_HB_TOOLS)/attribute_types_openpower.xml \
+            $(MRW_HB_TOOLS)/attribute_types_oppowervm.xml \
```
williamspatrick: Should we change these to simply be wildcards, so we don't have to keep updating them?

$(wildcard $(MRW_HB_TOOLS)/attribute_types*.xml) \
$(wildcard $(MRW_HB_TOOLS)/target_types*.xml)

e-liner:
Yeah, I was thinking that when I made this commit.  I'm not sure if there will be any other changes now that we have these two in place, but it's definitely possible.   Plus the wildcard is a lot more manageable.  I'll change it.

dcrowell77:
A wildcard defeats the purpose of this support, but the original change is missing a check as well.  The whole point of those extra files is that we want to conditionally include them in the compile and build of the binary.  I believe the correct version of these changes is to pull in the generated xml files from the hostboot build that has (or has not) pulled in the additional xml files.

e-liner:
Right. The hostboot side of things has the configuration flags to decide whether or not a file is compiled and if it is, op-build pulls it in.  Or are you saying that the op-build side of things needs an additional config flag?

dcrowell77:
I don't think op-build should know anything about any of these extra files.  The hostboot package should create a common/consistent set of files that are consumed by the other package(s), reflecting what the CONFIG flags inside Hostboot say.

stewart-ibm:
retest this please

e-liner:
retest this please

e-liner:
retest this please

stewart-ibm:
@e-liner something you still want to merge? There's conflicts and I haven't seen it build successfully yet :(

stewart-ibm:
No activity for a long while, so I'm closing this pull req.

Feel free to re-open if this is something that's wanted.

PR #631, opened 2016-09-12
Update hostboot master-p8 commit

bofferdn:
- Point to commit which supports secureboot CRTM verification

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/631)


williamspatrick:
We have a tool that automatically does this. We done need these sort of commits for most of the repos.

PR #634, opened 2016-09-12
Remove p9 hostboot patch (unused variable)

ploetzma:
Fix was included with https://github.com/open-power/op-build/pull/632

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/634)


ploetzma:
@jazurin I think there was a part of your pull request script that checked if the new hostboot code included any patch fixes. Does that have to be fixed up for p9?

jazurin:
@ploetzma that should be taken care of already.  For P9 the script looks in the 'p9Patches' dir, and for P8 in 'p8Patches'.  If the gerrit changeId of a patch is found in the new hostboot commit history, then that patch is removed as part of the PR.

ploetzma:
@jazurin Ah, understood. Looks like my change-id changed in some path. Thanks for the reply.

PR #637, opened 2016-09-14
scripts: Drop support for CAICOS cards

sammj:
The CAICOS series of radeon GPUs work on POWER but do not appear to
survive kexec. Remove support for them from Skiroot so that the GPU
won't be properly initialised until the host OS boots.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/637)


sammj:
@stewart-ibm this should also go to op-build-1.10.y

PR #642, opened 2016-09-14
Support sb-signing-utils package

bofferdn:
- Added patch containing highly experimental signing utilities
- Added configuration file with option to configure package
- Added public HW + SW keys
- Added package makefile
- Registered package with buildroot

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/642)


bofferdn:
Replaces https://github.com/open-power/op-build/pull/640

bofferdn:
I found a problem with Sub's code today where the the order of initialization of some strings is causing it to fail.  It was a crap shoot that I didn't hit it earlier.  I'll need to get a patch from him.

stewart-ibm:
Note: there's a lot of activity going on in review of the tools themselves.

Previous comments on https://github.com/open-power/op-build/pull/640 still apply, including "we should do normal things and fetch it from a source tarball, or if we must, from git."

stewart-ibm:
Review of the tool itself is over at: https://github.com/open-power/sb-signing-utils/pull/1

Although there hasn't been any response in >10 days, so having it make 1.12 is looking... unlikely.

bofferdn:
This is the replacement of pull 640

bofferdn:
whoops wrong pull request.

bofferdn:
Replaced with pull request #677

PR #627, opened 2016-09-07
Release 4.4.20-openpower1

shenki:
v4.4.20 stable release, which includes a fix for a NVMe issue that is
important for OpenPower machines.

Signed-off-by: Joel Stanley <joel@jms.id.au>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/627)

PR #646, opened 2016-09-15
Linux 4.4.21 openpower1

shenki:
This pull request incorporates the 4.4.20-openpower1 kernel as well as the 4.4.21-openpower1 commit. The 4.4.20 release was not merged due to an unrelated CI failure, however, it makes sense to include it in our history should we want to do regression testing.

v4.4.21 stable release includes a fix for the fix for the NVMe issue that is important for OpenPower machines.

Replaces #627.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/646)


williamspatrick:
Did we lose a config value?  We didn't end up with a zimage.papr file.

shenki:
```
>>> linux 4.4.21 Downloading
--2016-09-15 06:02:37--  https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-4.4.21.tar.xz
Resolving cdn.kernel.org... 151.101.68.69
Connecting to cdn.kernel.org|151.101.68.69|:443... connected.
ERROR: certificate common name “k.ssl.fastly.net” doesn’t match requested host name “cdn.kernel.org”.
To connect to cdn.kernel.org insecurely, use ‘--no-check-certificate’.
```

Same bug as https://github.com/open-power/op-build/pull/627. wget is too old.

williamspatrick:
Can we solve this with either:

1. Adding --no-check-certificate
2. Depending on host-wget

I absolutely hate (1) from a security stand-point.  Maybe there is something we can do for the kernel and only if wget is a certain age?

All of our developers internally are using RHEL6, so we need to find a way to support building on that. Otherwise we will have made it so that the entire internal team cannot do their job.  Having a custom wget that they add to their path isn't the best solution.

williamspatrick:
Should we set a custom URL to point to our github tree's tar file instead?

stewart-ibm:
> Can we solve this with either:
>
> 1. Adding --no-check-certificate
> 2. Depending on host-wget
>
> I absolutely hate (1) from a security stand-point.  Maybe there is
> something we can do for the kernel and only if wget is a certain age?

I'd go for 2 rather than 1... although I'm not sure the buildroot magic
to do that.

-- 
Stewart Smith
OPAL Architect, IBM.

shenki:
There's no such thing as host-wget in buildroot (I had a look for it).

There's some action happening on https://github.com/open-power/op-build/pull/627 to get this fixed for CI.

jazurin:
added  to linux-4.4.21.tar.xz to the local download cache.
retest this please

PR #671, opened 2016-10-03
Fix the proxy issue in ci

sword-huang:
The proxy setting should add into environment variable, too!

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/671)

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -60,16 +60,15 @@ do
 	fi
 	if [[ -n "$http_proxy" ]]; then
 	  if [[ "$distro" == fedora23 ]]; then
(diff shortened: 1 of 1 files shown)


stewart-ibm:
ok to test

williamspatrick:
@sword-huang - Why did you close this?

sword-huang:
I find if I use normal user to execute the script, it can work. It doesn't
need to be modified.
Sorry for your trouble.

Jason Huang

2016-10-03 19:47 GMT+08:00 Patrick Williams <notifications@github.com>:

PR #675, opened 2016-10-06
Add p9 scale out simulation build support

ploetzma:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/675)


openpower/package/p9so_sim-xml/p9so_sim.mk
```diff
@@ -0,0 +1,78 @@
+################################################################################
+#
+# p9so_sim_xml
+#
+################################################################################
+
+P9SO_SIM_XML_VERSION ?= a89100084400694a585273e64f9696b6f3b4da48
+P9SO_SIM_XML_SITE ?= $(call github,sannerd,p9so_sim-xml,$(P9SO_SIM_XML_VERSION))
```
ploetzma: Will update these when p9so__sim xml repo is in

williamspatrick:
Why are we doing this?  Is there anyone outside of IBM that has access to this simulator?

ploetzma:
@williamspatrick It sounds like this will be utilized by the HDAT team
@sannerd Is that correct?

dcrowell77:
@stewart-ibm Asked for it I think

sannerd:
@williamspatrick  This is desired to allow easier simulation between various teams.

stewart-ibm:
So, I think the Jenkins tests should build all defconfigs, as otherwise we have situations like this one, where it tests okay, but doesn't actulaly test that the new code builds. @e-liner is this something you can look at?

stewart-ibm:
As for having something internal or external, we had the P9 Mambo defconfig for months before there was ever a public P9 Mambo, so I don't see it as a real problem - but if there's a point where it's no longer going to be maintained or tested, then it should drop.

openpower/package/p9so_sim-xml/p9so_sim.mk
stewart-ibm: Of course, should be an open-power repo before we merge :)

stewart-ibm:
Okay... we should make a call on this.

Is this going to give us anything in the next X few months that the zaius/witherspoon builds do not?

ploetzma:
@sannerd Looking for your input here. Is there a p9so__sim xml repo in the works?

PR #677, opened 2016-10-06
Support sb-signing-utils package

bofferdn:
- Added configuration file with option to configure package
- Added public HW + SW keys
- Added package makefile
- Registered package with buildroot
- Added several pre-req perl modules

---

This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/677)


bofferdn:
superscedes pull request #642

stewart-ibm:
Please explain what has changed since previous request.

You don't need a new pull request when you rebase/add commits, github will automatically deal with it.

stewart-ibm:
I've marked in-progress as the tool itself is not yet ready/reviewed with review comments addressed.

bofferdn:
The changes include:
a) Strips out any/all "patches" , as the pull request for sb-signing-utils should have everything
b) It notably does not move the project pointer (currently only a LICENSE file in the sb-signing-utils repo)
c) Fixes up the makefile to use the native autotools support, making everything a lot simpler
d) Adds additional pre-req host perl packages (4) that will be used by sb-signing-utils scripts once production signing functionality is vetted
e) Fixes up the text in the private keys, adds a README discussing aspects of secureboot, how the keys are related, and manufacturer expectations on those keys

stewart-ibm:
Why are the key files 133 bytes yet the container header field is 132 bytes?

stewart-ibm:
Considering https://github.com/open-power/sb-signing-utils/pull/1 hasn't gotten any closer to being mergable, I'm closing this pull request.

It looks like there isn't really an effort to get the generation of secure/trusted boot headers for p8 (and I'm guessing P9) into upstream code, which is sad. I was hoping we could have op-build generating stb headers across the board, and be at the very least verifying them in software on all platforms by now.

PR #708, opened 2016-10-26
Turn on console, AST_2500 for Witherspoon

ploetzma:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/708)


shenki:
Not ast2500?

williamspatrick:
@ploetzma - Please put a change into hostboot such that AST2500 is used as the name instead, even if it is simply an alias to AST2400 so that we stop getting this line of questioning and place that change also as a patch here.

ploetzma:
@williamspatrick Updated patch with AST2500 alias to AST2400

PR #714, opened 2016-10-31
ppe42-gcc package update 10/31

dgilbert999:
Point to the latest ppe42-gcc package

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/714)


williamspatrick:
Merging even though CI failed because there must be some general CI problem.  This one failed on Habanero which doesn't even use this package.

PR #719, opened 2016-11-02
Update witherspoon.config with bring up options

ploetzma:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/719)


openpower/package/hostboot/p9Patches/hostboot-0004-include-pvpd.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/src/usr/vpd/dvpd.C b/src/usr/vpd/dvpd.C
+index 9ce0999..2839172 100644
+--- a/src/usr/vpd/dvpd.C
++++ b/src/usr/vpd/dvpd.C
+@@ -43,6 +43,7 @@
+ #include "dvpd.H"
+ #include "cvpd.H"
+ #include "vpd.H"
++#include "pvpd.H"
```
stewart-ibm: Is this hitting upstream hostboot? Why should we carry this as a patch?

stewart-ibm:
@e-liner can we have witherspoon added to the CI build?

openpower/package/hostboot/p9Patches/hostboot-0004-include-pvpd.patch
ploetzma: Yes, this is in review for upstream Hostboot. We are carrying it as a patch for now because without this change, the set PVPD_WRITE_TO_HW config does not compile.

openpower/configs/hostboot/witherspoon.config
```diff
@@ -1,48 +1,46 @@
...
 set MVPD_READ_FROM_PNOR
 set MVPD_WRITE_TO_PNOR
-unset DJVPD_READ_FROM_HW
+set DJVPD_READ_FROM_HW
 unset DJVPD_WRITE_TO_HW
 set DJVPD_READ_FROM_PNOR
 set DJVPD_WRITE_TO_PNOR
-unset CVPD_READ_FROM_HW
-unset CVPD_WRITE_TO_HW
+set CVPD_READ_FROM_HW
```
williamspatrick: This is, on the surface, kind of odd to see for Witherspoon.  There is no Centaur and thus no CVPD for Witherspoon.

I suspect there is some odd arrangement between CVPD and MVPD code though that requires this.

openpower/configs/hostboot/witherspoon.config
```diff
@@ -1,48 +1,46 @@
 # The Serial Flash Controller is the AST2400 BMC.
-set   SFC_IS_AST2400
+set   SFC_IS_AST2500
+unset SFC_IS_AST2400
```
williamspatrick: Why do we need to unset this?  Why is it set by default?

openpower/configs/hostboot/witherspoon.config
dcrowell77: You are correct.  The 'CVPD' name has been re-used to handle all of the planar VPD (that used to also contain the centaur/membuf info).  Someday maybe we'll fix up the names...

openpower/configs/hostboot/witherspoon.config
williamspatrick: 'SFC_IS_AST2400' should be 'default if !SFC_IS_AST2500' and not 'default y'.

ploetzma: I will make the change and add a patch to this commit

dcrowell77: The default comes from the HB repo though.  So are we going to hold off on this change until the HB change gets merged through?

PR #756, opened 2016-11-18
Updates to generate witherspoon pnor successfully

pragupta:
- Update witherspoon.mk to use processMrw and point to the latest xml
    - Update hostboot's witherspoon.config
    - Add patches for processMrw and dvpd fix

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/756)


e-liner:
ok to test

PR #771, opened 2016-12-08
Load 'sg' kernel module automatically in petitboot

mamatha4:
This patch is to add udev rule to load "sg" kernel module
automatically when petitboot is up.

Signed-off-by: Mamatha Inamdar <mamatha4@linux.vnet.ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/771)


sammj:
For context this enables the USB interface to work automatically for ipmitool and/or update_flash, without having to modprobe the module.

That looks like it's failing just on Witherspoon - master-next has that fixed, and this should probably be based on that anyway. @mamatha4 could you make this PR against master-next instead?

mamatha4:
Created new PR against master-next
 Load 'sg' kernel module automatically in petitboot #772

PR #789, opened 2016-12-23
overlay: Move `udevadm settle` to after pb-discover start

sammj:
Add S10udev and S17udevsettle to the overlay to replace the existing
S10udev from buildroot. `udevadm settle` can wait up to 30 seconds which
delays the start of pb-discover and can cause important information to
miss the Petitboot UI. Start pb-discover first and then settle to have
the UI receive all events and bring up devices in the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/789)

### DIFF
--- a/openpower/overlay/etc/init.d/S10udev
+++ b/openpower/overlay/etc/init.d/S10udev
@@ -1,0 +1,49 @@
+#!/bin/sh
+#
+# udev	This is a minimal non-LSB version of a UDEV startup script.  It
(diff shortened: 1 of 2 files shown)


sammj:
This is very open for bikeshedding, mainly up as an example of how to solve the perceived startup delay issue in Petitboot. This could also be done by patching package/eudev in our buildroot tree.

PR #793, opened 2017-01-02
openpower/packages: Add a new package

maddy-kerneldev:
Initial commit to add support to build
CATALOG for In Memory Accumulation/Counter [IMA].

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/793)


maddy-kerneldev:
We also have a pnor changes needs to be pulled alone with this pull request. 
https://github.com/open-power/pnor/pull/54/commits/fb8f459cddea5113dc96ca61058565f07650d817

e-liner:
We'll have to wait until that pull request is merged in the pnor repository to retest this, because I'm assuming that these changes depend on that.

ploetzma:
Should you also be removing the IMA catalog binaries in the hostboot-binaries since your change will make them unused?

PR #804, opened 2017-01-11
Release Linux 4.9.2-openpower1

shenki:
This moves from the 4.4 based kernel series to the newer long term
support kernel, Linux 4.9.

To make the move we regenerate the kernel defconfig for the new kernel.

We still have three out of tree patches. This commit makes the following
changes:

  * Drops 5 of our out of tree patches
  * Two remain for XHCI kexec issues
  * One remains for VGA handover

We also update bnx2x firmware version for newer kernel driver, which
means modify the firmware whitelist.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/804)


stewart-ibm:
retest this please

stewart-ibm:
Also, why on earth are we still hardcoding platforms in jenkins? Not everything is going to have all platforms, so it just serves to break things.

stewart-ibm:
retest this please

stewart-ibm:
retest this please

stewart-ibm:
(zauis us been merged, so this should pass now)

PR #806, opened 2017-01-12
configs: Update BR2_EXTERNAL usage.

sammj:
Commit 5ab5fdb "Update to Buildroot 2016.11" updated each defconfig to
use the new BR2_EXTERNAL convention, but the recently added lines for
BR2_PACKAGE_BUSYBOX_CONFIG_FRAGMENT_FILES were missed. Update these to
correctly use BR2_EXTERNAL_OP_BUILD_PATH instead.

Signed-off-by: Samuel Mendoza-Jonas <sam@mendozajonas.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/806)

PR #809, opened 2017-01-14
ZZ defconfig

stewart-ibm:
Similar to firenze, but more recent versions of all the things.

It should be relatively close to booting, but not quite yet until we bump the skiboot revision again after merging more HDAT patches.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/809)


openpower/configs/zz_defconfig
```diff
@@ -0,0 +1,56 @@
...
+BR2_GENERATE_LOCALE="en_US.UTF-8 de_DE.UTF-8 es_ES.UTF-8 fr_FR.UTF-8 it_IT.UTF-8 ja_JP.UTF-8 ko_KR.UTF-8 pt_BR.UTF-8 ru_RU.UTF-8 zh_CN.UTF-8 zh_TW.UTF-8"
+BR2_ROOTFS_OVERLAY="../openpower/overlay"
+BR2_ROOTFS_POST_BUILD_SCRIPT="../openpower/scripts/fixup-target-var ../openpower/scripts/firmware-whitelist"
+BR2_LINUX_KERNEL=y
+BR2_LINUX_KERNEL_CUSTOM_VERSION=y
+BR2_LINUX_KERNEL_CUSTOM_VERSION_VALUE="4.10-rc3"
+BR2_LINUX_KERNEL_USE_CUSTOM_CONFIG=y
+BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE="$(BR2_EXTERNAL_OP_BUILD_PATH)/configs/linux/skiroot_defconfig"
+BR2_LINUX_KERNEL_ZIMAGE_EPAPR=y
+BR2_PACKAGE_BUSYBOX_CONFIG_FRAGMENT_FILES="$(BR2_EXTERNAL)/configs/busybox.fragment"
```
sammj: This should be BR2_EXTERNAL_OP_BUILD_PATH (the rest being fixed up by https://github.com/open-power/op-build/pull/806)

williamspatrick:
Are we thinking of adding these for every single machine?  What is the decision process on when there is a defconfig and when there is not?  It is a little confusing to me, if I were to know nothing, as to why there is no hostboot for the zz machine.

stewart-ibm:
> Are we thinking of adding these for every single machine?  What is the
> decision process on when there is a defconfig and when there is not?
> It is a little confusing to me, if I were to know nothing, as to why
> there is no hostboot for the zz machine.

We should probably start to have some documentation around the
defconfigs - and perhaps around creating your own for a new platform.

I'm of the opinion that we accept any-and-all ones for any system that
somebody both wants to ship and will spend some effort helping to
maintain functionality of.

In this case, it does simplify complying with the GPL requirement of
providing "complete corresponding source code".

-- 
Stewart Smith
OPAL Architect, IBM.

Issue #803, opened 2017-01-10
ppe42-binutils doesn't build with recent GCC (e.g. on Fedora 25)


stewart-ibm:
With GCC6:
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I. -I. -I./../include -DHAVE_bfd_elf32_powerpc_vec -DHAVE_rs6000coff_vec -DHAVE_bfd_elf32_powerpcle_vec -DHAVE_ppcboot_vec -DHAVE_bfd_elf64_powerpc_vec -DHAVE_bfd_elf64_powerpcle_vec -DHAVE_bfd_elf64_little_generic_vec -DHAVE_bfd_elf64_big_generic_vec -DHAVE_bfd_elf32_little_generic_vec -DHAVE_bfd_elf32_big_generic_vec -DBINDIR=\"/home/stewart/op-build/output/host/usr/powerpc64le-buildroot-linux-gnu/sysroot/ppe42-binutils/linux/bin\" -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -Werror -Wno-error=unused-value -MT elf64-ppc.lo -MD -MP -MF .deps/elf64-ppc.Tpo -c elf64-ppc.c  -fPIC -DPIC -o .libs/elf64-ppc.o
elf64-ppc.c: In function ‘ppc64_elf_size_stubs’:
elf64-ppc.c:12257:16: error: left shift of negative value [-Werror=shift-negative-value]
          & (-1 << htab->plt_stub_align));
                ^~
elf64-ppc.c: In function ‘ppc64_elf_build_stubs’:
elf64-ppc.c:12725:12: error: left shift of negative value [-Werror=shift-negative-value]
      & (-1 << htab->plt_stub_align));
            ^~
cc1: all warnings being treated as errors
Makefile:1612: recipe for target 'elf64-ppc.lo' failed
make[6]: *** [elf64-ppc.lo] Error 1
Makefile:1654: recipe for target 'all-recursive' failed
make[5]: *** [all-recursive] Error 1
Makefile:1138: recipe for target 'all' failed
make[4]: *** [all] Error 2
Makefile:2598: recipe for target 'all-bfd' failed
make[3]: *** [all-bfd] Error 2
Makefile:829: recipe for target 'all' failed
make[2]: *** [all] Error 2
[...]

PR #825, opened 2017-01-19
Add support for IMA-CATALOG package

ploetzma:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/825)


williamspatrick:
Need Zaius updated.

ploetzma:
Zaius defconfig has been updated.

e-liner:
@williamspatrick Could you take a look at this?  op-build pull requests will continue to fail until it's merged.

rlippert:
FYI it looks like this is breaking builds for me since there is no build time dependency between openpower-pnor and ima-catalog packages, so the openpower-pnor build complains it can't find the ima-catalog.bin file.  I think OPENPOWER_PNOR_DEPENDENCIES needs to be updated to add ima-catalog package as (optional) dependency?

stewart-ibm:
Thanks to https://github.com/open-power/op-build/issues/803 I have issues building any P9 system on machines I have readily available, so debugging may be slower.

@ploetzma are you able to jump in here?

ploetzma:
I have helped @wghoffa through this issue, I believe you are describing it correctly. I will update the dependencies to include ima-catalog.

PR #864, opened 2017-02-11
ppe42-gcc: Host package should depend on host-ppe42-binutils

wak-google:
If host-ppe42-gcc is built against a buildroot version of 2016.11
then the dependency of ppe42-binutils is ignored as it is not
specified as a host dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/864)

### DIFF
--- a/openpower/package/ppe42-gcc/ppe42-gcc.mk
+++ b/openpower/package/ppe42-gcc/ppe42-gcc.mk
@@ -9,6 +9,7 @@ PPE42_GCC_SITE ?= $(call github,open-power,ppe42-gcc,$(PPE42_GCC_VERSION))
 PPE42_GCC_LICENSE = GPLv3+
 
 PPE42_GCC_DEPENDENCIES = ppe42-binutils
+HOST_PPE42_GCC_DEPENDENCIES = host-ppe42-binutils
 
 PPE42_GCC_DIR = $(STAGING_DIR)/ppe42-binutils
 PPE42_GCC_BIN = $(STAGING_DIR)/ppe42-binutils/linux


stewart-ibm:
ok to test

williamspatrick:
@e-liner - Was there a problem with CI over the weekend?  I'm going to assume the HW fails are not related to this commit.

williamspatrick:
@shenki - Is this a general pattern we need to fix up now?  I don't understand why they would have removed the behavior in Buildroot to automatically take package deps and apply them to HOST if building the host variant.

e-liner:
Yeah, I'm looking into it. I think one of our machines needs a reboot.

shenki:
@williamspatrick I'm not aware of any change

williamspatrick:
@wak-google - Was this a one-off oversight or was this really a buildroot 2016.11-based fix?  I'm trying to figure out if we have a bigger problem.

wak-google:
My understanding when sending changes upstream is that they expect _DEPENDENCIES to be specified explicitly for the HOST and TARGET builds. However, reading the current codebase for the generic builder it looks like it should do the conversion of TARGETS deps to HOST deps like it always did in the past. In theory if this is a real issue it would affect more than just this package. I had issues when building the occ for zaius because nothing was depending on the ppe42-binutils package. Once I applied this change the dependencies were correct. I'll try and dig more into this.

My guess is that may it has something to do with the fact that ppe42-gcc and ppe42-binutils are host-only packages and therefore never pull in the target deps?

e-liner:
retest this please

wak-google:
Any updates on fixes for this?

stewart-ibm:
I also asked one of our buildroot gurus about it, looks good to him, so looks good to me. Merged.

PR #868, opened 2017-02-13
P9 bu 17 02 06.rel

sannerd:
Get the base P9 BU stuff into master-next

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/868)


openpower/package/hostboot-binaries/hostboot_binaries.mk
```diff
@@ -4,8 +4,8 @@
 #
 ################################################################################
 
-HOSTBOOT_BINARIES_VERSION ?= 7425fb1d540920f0397f57f6c54a180f4f6a720e
-HOSTBOOT_BINARIES_SITE ?= $(call github,open-power,hostboot-binaries,$(HOSTBOOT_BINARIES_VERSION))
+HOSTBOOT_BINARIES_VERSION ?= 74bff2b31ba2328d9caee6e8ce2d6b6c9db654fc
+HOSTBOOT_BINARIES_SITE ?= $(call github,sannerd,hostboot-binaries,$(HOSTBOOT_BINARIES_VERSION))
```
williamspatrick: We can't pick this up into op-build.  Need a real repository to point at.

openpower/package/hostboot/hostboot.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 HOSTBOOT_VERSION_BRANCH_MASTER_P8 ?= f2b28bf6eee57757be45c45e514d2542d72aa350
-HOSTBOOT_VERSION_BRANCH_MASTER ?= b9fd10a99329adb4093c453bf9f80b05a6f31fa3
+HOSTBOOT_VERSION_BRANCH_MASTER ?= eb217e512338e8d2182762cbe97f55d4759e799e
```
williamspatrick: I'm worried about seeing a version update here because the Hostboot team already has an auto-update job they run.  Is this really needed?

williamspatrick: I think their job automatically removes patches too, so the patch stuff worries me as well.

openpower/package/hostboot/p9Patches/hostboot-1000-fix-reboot-hang-when-data-in-hbel.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/src/usr/errl/errlmanager_common.C b/src/usr/errl/errlmanager_common.C
```
williamspatrick: Can't pick up patches that do not have Gerrit Change-IDs in them because it breaks the auto-update tool the Hostboot team has.

e-liner:
retest this please

stewart-ibm:
(I echo what @williamspatrick said)

I think it may be better to go down the route of this patchset doing the hostboot config changes, updating hostboot-binaries and adding in the extra hostboot patch while letting the auto hostboot update foo do the rest.

sannerd:
Working on getting the HB bins officially update and will work with  @wghoffa and the "auto" hostboot release

wghoffa:
@sannerd - This moves us up to everything merged at a certain point yesterday: https://github.com/open-power/op-build/pull/870/commits/fd6d0b66c18123c3a0ab2affd7042f75bfeb1eda

Your commit looks to be from 8 days ago, so we should be covered there.

sannerd:
@williamspatrick please re-review

openpower/package/hostboot-binaries/hostboot_binaries.mk
sannerd: Done

openpower/package/hostboot/hostboot.mk
sannerd: Done

openpower/package/hostboot/p9Patches/hostboot-1000-fix-reboot-hang-when-data-in-hbel.patch
sannerd: Done, found the bug and fixed

williamspatrick:
The hostboot-binaries are updated by an auto-update job.  There is a pull request for that already.

The zaius config change should be included here.

I am curious why the auto-update job didn't remove these patches already.  If they are merged through Hostboot that job should have removed them.  @e-liner ideas?

This errlcommon commit is so small that it should sail through Hostboot review.  I would prefer to get it merged for real rather than applying a patch for a few days.

sannerd:
@williamspatrick   Please review again.   

The Hostboot patches were dropped -- they never came through the Hostboot commit process and were fixed elsewhere (XML values for the mss_thermal, SBE inits for the IPOLL).

Issue #876, opened 2017-02-17
Hostboot and OCC P8/P9 split breaks VERSION


stewart-ibm:
no longer picks up patches as part of VERSION

dcrowell77:
@ploetzma - Please look into this.

stewart-ibm:
I have a fix. Basically, it's impossible to work out if p8Patches or p9Patches are being selected.

I think the correct way is to split it. Have a virtual 'hostboot' package with hostboot-p8 and hostboot-p9 providing it. Then, patches are discovered the normal way.

It also means we don't have funny if conditions around versions.

There's also *other* brokenness in the VERSION output that I have patches for (but haven't reported an issue yet).

If my build succeeds here, I'll post a pull request.

PR #877, opened 2017-02-17
Split P8 and P9 Hostboot packages

stewart-ibm:
This was causing a couple of issues:
1) increased complexity in build files
2) It breaks the pkg-version patch detection/checksum

The split makes this a bit easier and requires less custom code.
We have the p8 and p9 packages provide the 'hostboot' virtual package,
allowing all the existing dependencies to work.

Fixes: https://github.com/open-power/op-build/issues/876

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/877)


stewart-ibm:
(note: the conflict is just that the hostboot SHA has been updated since I made this patch. Trivial to resolve on merge.)

Yes, this may break the auto pull request scripts. I do **NOT** think that should block this.

dcrowell77:
@ploetzma @e-liner Please take a look

williamspatrick:
Are we going to do the same for OCC?

```
Yes, this may break the auto pull request scripts. I do NOT think that should block this.
```

Why not?  I don't see any rush to get this in and the affected team is certainly under a lot of pressure as it is with P9 work.

stewart-ibm:
> Are we going to do the same for OCC?

Yes.


> ```
> Yes, this may break the auto pull request scripts. I do NOT think that should block this.
> ```
>
> Why not?  I don't see any rush to get this in and the affected team is
> certainly under a lot of pressure as it is with P9 work.

It really helps with VERSION being sane.

(I went down the rabbit hole with Sam of trying to programatically
compare the version of current running firmware with one of a new
firmware version and work out if the new one really is newer... this is
one of the things that fell out of it all as kind of being needed.)

-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
@ploetzma @e-liner : We need to get this reviewed from your team's perspective.

williamspatrick:
@stewart-ibm:

1. Does the version script need updating too?
2. What do we intend to do with OCC?  @wilbryan FYI.

stewart-ibm:
> @stewart-ibm:
>
> 1. Does the version script need updating too?

No. With this split, it starts working correctly again (and this is all
independent of my recent patch to improve the logic of that script,
fd7a226a3c74aea01617148a8b13fd9d95320137 ).

VERSION has been broken r.e. Hostboot patch status since
3a184ddba3bca878cfd6e38879ab76451c0c95d1, which is op-build v1.12

Because the logic for picking p8Patches/p9Patches and applying them was
done by hand, the foo that the version script used to find patches
(looking in appropriate paths) no longer worked. So, I could add 1001
patches to hostboot, but the VERSION would show that I had 0 patches
applied (which is probably *not* what the hostboot team wants to see if
they get a bug report with a heavily patched hostboot)

> 2. What do we intend to do with OCC?  @wilbryan FYI.

Yes, we'll need to do the same thing for OCC.

-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
@stewart-ibm there is a list of packages in the 'release-notes' script though. We need that updated for the two new packages or am I remembering the code wrong?

stewart-ibm:
> @stewart-ibm there is a list of packages in the 'release-notes' script
> though. We need that updated for the two new packages or am I
> remembering the code wrong?

This bit of code does that work:
https://github.com/open-power/op-build/pull/877/files#diff-93d4cde30f297a9894dc11ed49f00a44R49

It adds the appropriate hostboot package for the openpower platform
(p8/p9) to the list of versioned ones.

-- 
Stewart Smith
OPAL Architect, IBM.

williamspatrick:
I think we are talking about two different things. openpower/scripts/release-notes

stewart-ibm:
> I think we are talking about two different
> things. openpower/scripts/release-notes

It may help if I properly read your question before answering it :)

Yes, that will likely need updating, should simplify/add clarity to the
initial p8/p9 split for it ( f4d24eea37935d1dd769243e555b35228dc265bf )
though.

-- 
Stewart Smith
OPAL Architect, IBM.

PR #879, opened 2017-02-17
ppe42-binutils,ppe42-gcc: Patch, use $(MAKE) and -Wno-error to fix bu…

stewart-ibm:
…ild on modern GCC

Because building a compiler with -Wno-error to avoid erorring out on
undefined behaviour is a good idea.

We build with $(MAKE) so that we get the -j option passed down from
buildroot. this speeds up build by a couple of orders of magnitude
on my build machine.

Fixes: https://github.com/open-power/op-build/issues/803

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/879)


stewart-ibm:
@dgilbert999 An ack from you for this one would be appreciated

dgilbert999:
These changes look good to me.

PR #886, opened 2017-02-23
Disable DRTM in Zaius config

bofferdn:
- Disable CONFIG_DRTM for Zaisu
- Disable CONFIG_DRTM_RIT_PROTECTION for Zaius

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/886)


bofferdn:
See 35633 hostboot gerrit commit; This commit breaks on Zaius due to config dependency (op-build config is pre-req to that commit)

williamspatrick:
Why doesn't this also affect Witherspoon?

bofferdn:
As far as I know .. the secureboot options have only been removed from Zaius by Dean.  A commit in HB is enabling some DRTM stuff, but Zaius only fails because the new config options for DRTM are dependent on SECUREBOOT + TPMDD flags being enabled which, for Zaius only, are unset.  So Zaius only breaks when we introduce that commit.

pragupta:
I think the reason why only zaius is breaking is because zaius has this:

unset SECUREBOOT
unset TPMDD
unset TPM_NUVOTON

as opposed to 

unset CONFIG_SECUREBOOT
unset CONFIG_TPMDD
unset CONFIG_TPM_NUVOTON

in witherspoon. We shouldn't have "CONFIG_" added to the beginning of the config flag. That is appended somewhere in the build flow. So, technically, you have the right configs in zaius and wrong in witherspoon. I thought I had a pull request to fix that at one point, but it didn't make it through.

bofferdn:
Resubmitted w/ #PR888 against master-next instead of master; included the fixes to Witherspoon.

PR #893, opened 2017-02-28
Moving back garrison version to fix hardware bug

e-liner:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/893)


stewart-ibm:
What bug? Why hasn't that commit been reverted in the garrison-xml repository?

e-liner:
There was a bug that caused our hardware to fail.   Erich Hauptli is working on a fix now, but to get us back to working, I went ahead and reverted the xml.

PR #910, opened 2017-03-07
Configure SPARSEMEM and NUMA for Petitboot Kernel

HankChang:
For Romulus or other P9 projects, it's probably best to have these features enabled in Petitboot kernel for end-users.

Signed-off-by: Hank Chang <chhank@tw.ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/910)


HankChang:
Otherwise, users might see the memory size information is incorrect and couldn't check the NUMA affinity or other details while in Petitboot.

PR #922, opened 2017-03-15
Fix conflict for Default to std=gnu++03 patch

scprek:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/922)

### DIFF
--- a/openpower/package/hostboot/p8Patches/hostboot-0003-Default-to-std-gnu-03.patch
+++ b/openpower/package/hostboot/p8Patches/hostboot-0003-Default-to-std-gnu-03.patch
@@ -1,26 +1,20 @@
-From bc7b8b21549752274b162c64dbce3c81d20f2cab Mon Sep 17 00:00:00 2001
-From: Stewart Smith <stewart@linux.vnet.ibm.com>
-Date: Thu, 25 Aug 2016 20:22:30 +1000
-Subject: [PATCH 03/10] Default to std=gnu++03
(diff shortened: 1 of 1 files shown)


dcrowell77:
Seems good to me

stewart-ibm:
Why not respin the original patch? It's still sitting waiting to be reviewed and merged into Hostboot, along with most of the others.

I also can't seem to find this change id (or anything else that would cause the conflict) in gerrit. URL?

dcrowell77:
I don't think these changes are compatible with the local build environment that we use for P8.  I'm investigating a way to make it dependent on the build environment.

scprek:
only hit our known ubuntu issues. Rebuilding now with the bad node down

stewart-ibm:
Stephen Cprek <notifications@github.com> writes:
> only hit our known ubuntu issues. Rebuilding now with the bad node
> down

I guess that's a known issue with something internal to HB team? I'm not
aware of any Ubuntu issue with upstream code.


-- 
Stewart Smith
OPAL Architect, IBM.

scprek:
@stewart-ibm Yes, only to our internal CI process. It's why the Jenkins test failed the first time.

scprek:
@dcrowell77 Looks like @stewart-ibm has implemented the patches in hostboot
https://ralgit01.raleigh.ibm.com/gerrit1/#/c/32209/6 No reviewers on it.

dcrowell77:
Yep, I pulled in a chunk of this request to see how it worked.

PR #947, opened 2017-03-20
POWER9 Hostboot: build attributeOverride if possible rather than never

stewart-ibm:
On ancient host compilers (GCC before 4.8), there isn't the -std=gnu++11
option available, which attributeOverride requires.

Instead of the original fix for this, which was to just never build it,
check to see if the host compiler supports it, and if so, build it.

Wanting Pam to look at this to check it does solve the issue she was having with building things for internal IBM consumption before we merge it.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/947)


stewart-ibm:
@ploetzma (if my patch works) This supersedes your patch. I think this may be a slightly more elegant way to solve it given the issues that Pam has been having with build machines running old gcc.

Thoughts?

dcrowell77:
Seems like we should just push the new code into hostboot directly instead of carrying a patch, then we can remove the old one when the new hostboot code shows up.  I did a local build (not op-build) and it didn't fail (also didn't produce binary as expected).  I have the patch going through hostboot CI now.

ploetzma:
@stewart-ibm @dcrowell77 I agree that the correct fix will be to have a way to check if the compiler supports c++11 and then build or do not build the attribute override.

The issue is for hostboot developers building locally, hostboot uses a fakeroot to compile the c++11 options. Otherwise, the /usr/bin/gcc version is 4.4.7 - too old. In my experience (and from what Dan states above) when you query compiler options during compile time it does not pull the fakeroot options, and thus we will not get an attribute override binary built (like we do currently).

    export FAKEROOT=${FAKEROOT:-/opt/mcp/shared/powerpc64-gcc-20150516}
    export CROSS_PREFIX=${CROSS_PREFIX:-${FAKEROOT}/wrappers/powerpc64-unknown-linux-gnu-}
    export HOST_PREFIX=${HOST_PREFIX:-${FAKEROOT}/wrappers/x86_64-pc-linux-gnu-}
    export PATH=${FAKEROOT}/wrappers:${PATH}****

stewart-ibm:
@dcrowell77  I git-review'd the code to get it into the HB CI, although there seems to be an odd failure or two (looks like the CI scripts try and build with the patch that's in op-build, which will obviously fail).

@ploetzma I used the exact compiler invocation used to build the utility, so it should work no matter what the setup, and do things based on the complier invocation that's going to be used for building.

fspbld:
I tried a build with this new patch and the binary doesn't build, so this patch is not quite right yet.

ploetzma:
@fspbld That was on the red hat machines that are currently building the witherspoon images?

fspbld:
@ploetzma  - yes

wghoffa:
@stewart-ibm - Are there some more changes coming from you? Are you looking for something from us?

stewart-ibm:
William Hoffa <notifications@github.com> writes:
> @stewart-ibm - Are there some more changes coming from you? Are you
> looking for something from us?

A couple of spare hours in the day to find the bug would be nice? :)

If somebody wants to fix this up that'd be great, otherwise I'll get
back to it post vacation, so around the end of April.

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
New revision of the patch seems to work correctly. Previously, I was erroniously including $(C1) in what I was testing for, this turns out being '@', which is not valid when executing shell.

fspbld:
Will try it out today.

stewart-ibm:
@fspbld @wghoffa So, the most recent revision of this patch should be good to go, but it'd be good if someone could confirm that thinks are working correctly everywhere (Pam, probably more directed at you :)

fspbld:
I can attempt again today to build with these changes.

fspbld:
@stewart-ibm  rh6-built op-build is OK (attributeOverride doesn't build, no errors seen). 
But, the binary is not built on rh7 (no errors, but looks like no attempt to build it was done).

fspbld:
More details emailed to Stewart for his review.

stewart-ibm:
okay.... after a day of broken CI, it looks like *this* time for sure.

I ended up having to produce a target no matter what, which on systems without a C++11 complier, will end up just being an empty shell script that does "exit -1" to indicate error. I think there's a bit of an oddity going on with some other rule in the hostboot makefiles I've not been able to quite figure out. Anyway, this current patch should work well enough for what's needed for the time being.

fspbld:
OK, will try it again  :)

ploetzma:
Pulled this in to test on local hostboot, and we still compile a local attributeOverride tool.
Looks good from my end.

ploetzma:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/open-power/op-build/947)*

PR #954, opened 2017-03-22
linux: Add p9 specific defconfig

sammj:
Add skiroot_p9_defconfig primarily so that we can add kernel command
line options only required for POWER9 systems. In particular
"powersave=off" is required while there are existing issues with cpuidle,
and "quiet" is removed to aid debugging during bringup.

While we're here also trim a few config options that are unlikely to be
required on P9:
	CONFIG_BE2ISCSI
	CONFIG_SCSI_SYM53C8XX_2
	CONFIG_SCSI_SYM53C8XX_DMA_ADDRESSING_MODE
	CONFIG_VORTEX
	CONFIG_PCNET32
	CONFIG_S2IO
	CONFIG_MYRI10GE

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/954)


sammj:
Please bikeshed whether there are more/less options to drop for P9! @mikey @shenki

PR #970, opened 2017-03-25
kernel: Fix cpuidle registration bug

svaidy:
Include kernel patch to pass correct drv->cpumask in
cpuidle-powernv driver.

This fix is required to boot on systems where cores are partially
marked as 'bad' in HDAT and further in skiboot device tree.

Without this fix, kernel will OOPS with
cpuidle_register_device passing incorrect memory addresses
to kobject init subsystem.

Upstream patch:
powerpc/powernv/cpuidle: Pass correct drv->cpumask for registration
http://patchwork.ozlabs.org/patch/742706/

Signed-off-by: Vaidyanathan Srinivasan <svaidy@linux.vnet.ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/970)


shenki:
Please take https://github.com/open-power/op-build/pull/971 instead of this one.

PR #983, opened 2017-03-27
Update ppe42-binutils version  ded0eff

dgilbert999:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/983)


stewart-ibm:
From looking at (internal) Slack, it seems this is needed so that SBE build using op-build compilers will work. Merging.

PR #999, opened 2017-04-03
Release Linux v4.10.7-openpower1

shenki:

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/999)


stewart-ibm:
retest this please

stewart-ibm:
@e-liner I still can't see any test results there:
```

Opening Robot Framework report failed

    Verify that you have JavaScript enabled in your browser.
    Make sure you are using a modern enough browser. Firefox 3.5, IE 8, or equivalent is required, newer browsers are recommended.
    Check are there messages in your browser's JavaScript error log. Please report the problem if you suspect you have encountered a bug.

```

I'm ignoring the P9 "failures" as they seem totally related to what's being run. All I see is a "SOL.log" that seems to just stop in the middle of skiboot booting, which could be an artifact of whatever test is being run not knowing about SBE updates, I'm really not sure, as every attempt to access any other log just gives me the above.

PR #1080, opened 2017-05-15
Fix up SECUREBOOT config flags

rjknight:
-removed CONFIG_ prefix to the config flags which were
     intended to disable secure boot

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1080)


bofferdn:
Jamie .. unless there is some overriding reason I am aware of, I think we should instead remove the existing 3 flags from the config entirely, so that secure boot remains active.  They want to be able to ship/test with a single driver image, therefore the driver of whether secure boot is enabled will be the jumper setup and other XML based policies.

rjknight:
removing the flags will disable secureboot until the corresponding changes are done in the hostboot code -  Should we just leave it as is until the config flags are removed from hosboot?

bofferdn:
actually what I meant was... the hostboot flags stay .. which default to secureboot compiled in; but for Witherspoon we'll remove the incorrect attempt to disable the compile.  In effect leaving it fully compiled in, but without the confusing attempt to disable it in the op-build config.

stewart-ibm:
Can one of the admins verify this patch?

stewart-ibm:

wghoffa:
@bofferdn @rjknight - Can we resolve this and close it if we aren't making changes?

PR #1088, opened 2017-05-18
HTMGT-OCC: Enable HTMGT compile, Load/Start OCC

baileysh:
RTC:171824

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1088)

PR #1096, opened 2017-05-19
Occ witherspoon plus skiboot 5.6.0 rc2

stewart-ibm:
skiboot bump plus https://github.com/open-power/op-build/pull/1088 which together should be enough to have functioning OCC on Witherspoon.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1096)


stewart-ibm:
retest this please

stewart-ibm:
Zaius machine is bricked, can't be used for test. Needs secondary SBE recovered.

The witherspoons seem to be just broken when it comes to doing expect to talk to the BMC.

So. I went and manually tested it on a Witherspoon, using op-test-framework. Things look good.

So, merging.

PR #1119, opened 2017-05-31
Enable BMC Mbox Protocol

wghoffa:
- This will enable Hostboot to use the BMC Mbox protocol to use
     the BMC to do PNOR accesses
   - Enable for Witherspoon + Zaius systems

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1119)


wghoffa:
retest this please

stewart-ibm:
No idea WTF was going on with witherspoon there, as far as I can tell, the machine booted fine.
So, ignore the CI "failure".

pragupta:
There was a false start, which puts the host in a quiesced state. We said this would be a legit failure. :)

PR #1149, opened 2017-06-14
Change AST and AACRAID drivers to be modules instead of built-in

guilhermepiccoli:
This patch changes the config files in order to make AST and AACRAID
drivers built as modules on Petitboot, instead of built-in.

Those drivers showed some issues lately that would've been
'workarounded' if both were modules, so we could blacklist
them. The total increase in BOOTKERNEL image is less than
80 kbytes.

Signed-off-by: Guilherme G. Piccoli <gpiccoli@linux.vnet.ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1149)


wghoffa:
retest this please

stewart-ibm:
Can one of the admins verify this patch?

wghoffa:
retest this please

PR #1177, opened 2017-06-22
Removing processMRW to unblock future changes

e-liner:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1177)


stewart-ibm:

PR #1179, opened 2017-06-23
Release Linux 4.11.6-openpower1

shenki:

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1179)


stewart-ibm:
The Zaius failure is a "known" one and not unique to this... the build test fail on my CI is as the result of a jenkins oddity (which is now fixed).

Otherwise looks good, merging!

PR #1214, opened 2017-07-05
Bump up HB, adjust OCC patches and fix HDAT I2C

sannerd:
- Latest HB has first in chain of WOF fixes
  - Existing patches to make OCC run fail to apply
  - Recent HDAT I2C commit only works on Witherspoon, so
    Add patch to allow other platforms to pass

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1214)

PR #1218, opened 2017-07-06
Bump up HB, adjust OCC patches and fix HDAT I2C

sannerd:
- Latest HB has first in chain of WOF fixes
  - Existing patches to make OCC run fail to apply
  - Recent HDAT I2C commit is causing issues everywhere,
    so add a patch to disable the I2C XML reporting

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1218)


sannerd:
Cleaner attempt to resolve the I2C xml issues on all systems from pull req https://github.com/open-power/op-build/pull/1214, etc

PR #1224, opened 2017-07-09
Remove OCC patch for DD2.0 to allow fix to apply

sannerd:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1224)


sannerd:
https://github.com/open-power/op-build/pull/1225 supersedes

PR #1230, opened 2017-07-10
witherspoon: rename to witherspoon-sequoia, witherspoon-redbud

stewart-ibm:
witherspoon: rename to witherspoon-sequoia, witherspoon-redbud
    
    Since Sequoia and Redbud planars currently require different images,
    and people are getting the wrong PNOR and flashing it, prevent that by
    forcing them to make a decision.
    
    Eventually this should go away, and we have one 'witherspoon'
    platform for both planars. Unfortunately, that day is not today.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1230)


williamspatrick:
Sequoia and Redbud are board names and not system names. Sequoia goes in Witherspoon and Redbud goes in something that looks kind of like a Witherspoon but doesn't have a name yet, right?  Why change the name to reflect a board name when we've always used system code names before?

dcrowell77:
Both Redbud and Sequoia are Witherspoon boards.  This is a gross violation of the way things should have gone, but the ship has sailed... sigh...

I'm not a huge fan of the name change either.  Don't builders already need to do jump through some hoops to pull in a fork of opal to make it work anyway?  How is this improving things?

@Erich-Hauptli @fspbld

fspbld:
Yeah, not sure I like the idea of the rename, either.  witherspoon just 'builds' for me now. For the redbud images we need (about once/week), I have to do that build manually.  Who are the people that are grabbing and installing the wrong image? Perhaps there is something that I can/should do to make it more clear which image should be installed. (I would prefer doing that over having to change the build/test/publish tools for a name change).

mikey:
Some arbitrary naming convention that no longer validly or logically applies and is causing problems, needs to be abandoned.
The idea of the rename is to force people to think about it before they hit a subtle issue.  Currently people get a witherspoon machine and grab the witherspoon.pnor and then wonder why it doesn't work. I've done this myself and had to help others with it.
Renaming to redbud and sequoia makes it clear which one is for what.
It's a violation of the naming convention but the issue that witherspoon.pnor doesn't work in all Witherspoons is the greater problem we need to fix.

mikey:
How about this alternative?  We still ditch the witherspoon.pnor name but call them witherspoon-sequoia and witherspoon-redbud (.pnor & _defconfig).  That way we keep the convention but still make it clear which is which.

dcrowell77:
I could live with witherspoon-sequoia and witherspoon-redbud for now I guess.

> the issue that witherspoon.pnor doesn't work in all Witherspoons is the greater problem we need to fix.
Preaching to the choir.

stewart-ibm:
> Both Redbud and Sequoia are Witherspoon boards.  This is a gross violation of the way things should have gone, but the ship has sailed... sigh...
>
> I'm not a huge fan of the name change either.  Don't builders already
> need to do jump through some hoops to pull in a fork of opal to make
> it work anyway?  How is this improving things?

That's now done by the redbud_defconfig, so that images can be built at
the same time, from same source and not alarm us with random OPAL patches.

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
> I could live with witherspoon-sequoia and witherspoon-redbud for now I
> guess.

Okay, that sounds like it'll make everybody the least unhappy, I'll do
that :)

>> the issue that witherspoon.pnor doesn't work in all Witherspoons is the greater problem we need to fix.
> Preaching to the choir.

I'm sure there'll be eight more names for it by the time it ships :(

-- 
Stewart Smith
OPAL Architect, IBM.

mikey:
@fspbld if you could make sure you honour these names when we do builds for public consumption (rather than renaming to plain witherspoon.pnor) I think that would help.

stewart-ibm:
(fixing up hostboot CI after merging)

PR #1247, opened 2017-07-14
Release notes bugfixes

stewart-ibm:
The witherspoon rename isn't as much fun as you'd think, and we have to jump through some hoops.

At the same time, this exposed a couple of bugs in the release-notes script.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1247)


sammj:
Tested and this produces a whole lot more than the "nothing" I previously got, so LGTM!

PR #1253, opened 2017-07-17
Add Base Witherspoon Files for Regression Testing to Work

wghoffa:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1253)


stewart-ibm:

stewart-ibm:
On intense promises in internal Slack (on pain of @apopple demanding photographs of each machine) I've been assured the machines in the test pool are all sequoia and made the adjustment to the Hostboot jenkins setup accordingly. This should all now pass.

We explicitly want to avoid having a witherspoon.pnor until there's a single witherspoon.pnor that does work across both planars.

PR #1256, opened 2017-07-18
Release 4.4.77-openpower1

shenki:
Update to the latest stable release. This pulls in some backports for
NVMe stability.

The "drivers/drm/ast: Switch SCU to VGA output on POST" patch was
forward ported so it applied.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1256)


stewart-ibm:
We have to ignore the Hostboot CI builds as they're buggy and hard code a list of platforms, not all of which were present back in op-build v1.10.x

PR #1280, opened 2017-07-27
release-notes: fix markdown output for tools like pandoc

stewart-ibm:

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1280)


stewart-ibm:
@williamspatrick yeah, mainly because I have no idea *why* or *where* this is documented as needed, but it does have the desired effect of when viewing HTML output from /usr/bin/pandoc or /usr/bin/markdown it much more closely reflects what github renders.

PR #1298, opened 2017-08-03
enable occ and enable check stop analysis

jimsmc:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1298)


stewart-ibm:

PR #1305, opened 2017-08-04
Witherspoon XML Patches for GPU Fixes

wghoffa:
- The witherspoon-xml package has regressions,
    so for now pull in the 2 commits we need as patches

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1305)


stewart-ibm:

wghoffa:
@stewart-ibm - Here is the downside with your machine-xml change. I can't think of a way to toss in XML patches? The problem here is witherspoon-xml/master is broken, but 2 commits between the level pointed to by op-build and witherspoon-xml/master are needed for an MFG driver.

scprek:
retest this please

scprek:
retest this please

stewart-ibm:
William Hoffa <notifications@github.com> writes:
> @stewart-ibm - Here is the downside with your machine-xml change. I
> can't think of a way to toss in XML patches? The problem here is
> witherspoon-xml/master is broken, but 2 commits between the level
> pointed to by op-build and witherspoon-xml/master are needed for an
> MFG driver.

(already mentioned in Slack, but reposting here for completeness)

https://github.com/open-power/op-build/tree/master/openpower/patches
that dir has patches that are specific to each platform
e.g. you can see the redbud patch for skiboot
and the machine-xml patches specific to habanero and palmetto
You also need something like this
https://github.com/open-power/op-build/blob/master/openpower/configs/witherspoon-redbud_defconfig#L3

that points buildroot to the dir for custom patches for that config.
so, we can, in fact, patch anything per defconfig.

so, for witherspoon, you'll need to add a BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_OP_BUILD_PATH)/patches/redbud-patches" style line to sequoia

and probably have the patch in both (or symlinked)
and then we get to resolve everything whenever we finally get a way to
programatically tell sequoia and redbud apart

-- 
Stewart Smith
OPAL Architect, IBM.

wghoffa:
@stewart-ibm - Thanks! Much appreciated.

PR #1324, opened 2017-08-10
Integrate sb signframework

hellerda:
This is a test patch only.  The "signframework" code is not yet in its permanent home, and I have some questions about build:

1. Is there a preference as to whether op-build packages are built as "generic" (simple makefile) or "autotools" (using GNU toolchain)?  The source code uses a simple makefile, so this patch uses "generic" for op-build integreation.  Hopefully this is adequate.  If it needs to be "autotools" we'll need to add GNU toolchain support to the project.

Remaining questions have to do with dependencies and "HOST" packages.  "signtool" has dependencies on openssl, libcurl and json.  Since all the signing utilities are HOST-only packages, the deps (I assume?) must be built as HOST packages also.  But upon digging into this, I realized I wasn't completely clear on how Include paths and RPATHs are supposed to work for HOST packages.

A TARGET package is bound for the image.  So, bits get installed to $(TARGET_DIR): e.g. $(TARGET_DIR)/usr/bin, $(TARGET_DIR)/usr/include and $(TARGET_DIR)/usr/lib.  For a HOST package, bits get installed to $(HOST_DIR)/usr/bin, $(HOST_DIR)/usr/include and $(HOST_DIR)/usr/lib.

A HOST package only needs to execute on the build machine.  This means that a HOST package may in fact compile if it's dependent headers are installed to /usr/include, and dependent libs are installed to /lib or /usr/lib.  But is this the correct way?  The questions are:

2. For a HOST package, is it acceptable to allow the build server's include path (i.e.. /usr/include) to provide the headers?  In which case, the -devel package may be installed on the build machine but not op-build?  Or *must* the includes be in $(HOST_DIR)/usr/include?  In which case, the dependency needs to built as a HOST package in op-build?

3. Same questions reqarding library path: for a HOST package: is it acceptable to allow the build server's library path (e.g. /usr/lib) to provide the libs?  Or *must* the libs be in $(HOST_DIR)/usr/lib?

There seems to be a job in buildroot that checks the RPATH of any file installed to $(HOST_DIR)/usr/bin:

```
$ op-build host-signframework-rebuild
...
>>> host-signframework master Installing to host directory
cp -p /home/build/op-build/master/op-build/output/build/host-signframework-master/src/client/client /home/build/op-build/master/op-build/output/host/usr/bin/
chmod 755 `/home/build/op-build/master/op-build/output/host/usr/bin/`
***
*** ERROR: package host-signframework installs executables without proper RPATH:
***   /home/build/op-build/master/op-build/output/host/usr/bin/client

```
You will get the above error if the RPATH does not look like this:

```
$ readelf -d output/build/host-signframework-master/src/client/client | egrep 'RPATH|RUNPATH'
 0x000000000000000f (RPATH)              Library rpath: [/usr/lib:/home/build/op-build/master/op-build/output/host/usr/lib]
```

Other paths *may* be in the RPATH, but it *must* include $(HOST_DIR)/usr/lib.

But then, in building some executables I see, the actual library path may be somewhere other than $(HOST_DIR)/usr/lib.  For example:

```
build@ltc254:~/op-build/master/op-build$ ldd output/build/host-signframework-master/src/client/client
        linux-vdso.so.1 =>  (0x00007ffc55fdb000)
        libjson-c.so.2 => /home/build/op-build/master/op-build/output/host/usr/lib/libjson-c.so.2 (0x00007f609716b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6096e59000)
        libcurl.so.4 => /usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f6096be9000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f609681f000)
        libidn.so.11 => /usr/lib/x86_64-linux-gnu/libidn.so.11 (0x00007f60965ec000)
        librtmp.so.1 => /usr/lib/x86_64-linux-gnu/librtmp.so.1 (0x00007f60963ce000)
        libssl.so.1.0.0 => /lib/x86_64-linux-gnu/libssl.so.1.0.0 (0x00007f6096165000)
        libcrypto.so.1.0.0 => /lib/x86_64-linux-gnu/libcrypto.so.1.0.0 (0x00007f6095d23000)
        ...
        libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007f6091724000)
```

So, apparently not every linkable module is expected to be in $(HOST_DIR)/usr/lib.  What's the rule here?

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1324)


hellerda:
Almost forget to mention: My patch assumes the json-c and libcurl must be build as HOST packages.  If that turns out to not be the case, this can be ignored.  If they *do* need to be built as HOST packages, please advise me here.

My .mk file has the following.  These are both "buildroot" packages:

```
+HOST_SIGNFRAMEWORK_DEPENDENCIES += host-json-c
+HOST_SIGNFRAMEWORK_DEPENDENCIES += libcurl
```

For json-c:  There is apparently no support to build this as a HOST package, so I made it work by adding the following to "json-c.mk".  Then the above will work, and the package does not need to be installed on the build server.  This is a patch to buildroot:

```
diff --git a/package/json-c/json-c.mk b/package/json-c/json-c.mk
index 7adbcaf..a8f23d2 100644
--- a/package/json-c/json-c.mk
+++ b/package/json-c/json-c.mk
@@ -14,4 +14,15 @@ JSON_C_AUTORECONF = YES
 JSON_C_LICENSE = MIT
 JSON_C_LICENSE_FILES = COPYING
 
+HOST_JSON_C_VERSION = json-c-0.12.1-20160607
+HOST_JSON_C_SITE = $(call github,json-c,json-c,$(JSON_C_VERSION))
+HOST_JSON_C_INSTALL_STAGING = YES
+HOST_JSON_C_MAKE = $(HOST_MAKE_ENV) $(MAKE1)
+HOST_JSON_C_CONF_OPTS = --disable-oldname-compat
+# AUTORECONF is needed because of Makefile.am.inc patch.
+HOST_JSON_C_AUTORECONF = YES
+HOST_JSON_C_LICENSE = MIT
+HOST_JSON_C_LICENSE_FILES = COPYING
+
 $(eval $(autotools-package))
+$(eval $(host-autotools-package))
```

For libcurl:  Apparently that *does* build as host package... at least, some bits get installed to $(HOST_DIR)/... even though it's not named "host-*"  However, it installs the header files to a rather inconvenient location, and I needed to do the following to make it work:

`(cd output/host/usr/include/; ln -s ../powerpc64le-buildroot-linux-gnu/sysroot/usr/include/curl curl)
`
Surely there is a proper way to do this but I wasn't clear on how.  But with this, the above will work, and the package does not need to be installed on the build server.

Please advise whether these "buildroot" packages need these sort of changes, or if there is a better way.

hellerda:
Oops, accidentally closed it.  Now I know what "Close and comment" means.  ;- )  Reopening...

hellerda:
I just opened pull request #1384 to replace this.  Closing this pull request.

PR #1327, opened 2017-08-11
Adding support for MEMD processing

e-liner:
This commit adds in the call to the perl script housed in
the hostboot build files.  It generates the MEMD header from
the directory of files given, compiles them into one file and
returns the output where it can be added to the PNOR.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1327)


stewart-ibm:

Issue #1194, opened 2017-06-28
hostboot patches fail silently


shenki:
When a patch in the openpower/packages/hostboot/hostbootP9 directory fails to apply, the build continues silently without applying further patches.

Instead it should fail when applying the patch, as something has gone wrong.

stewart-ibm:
Joel Stanley <notifications@github.com> writes:
> When a patch in the openpower/packages/hostboot/hostbootP9 directory fails to apply, the build continues silently without applying further patches.
>
> Instead it should fail when applying the patch, as something has gone
> wrong.

The best way to fix this is probably to split hostboot (and occ) out
into hostboot-p8 and hostboot-p9, that way we'll use the normal
buildroot patching rather than the custom stuff in there now.

-- 
Stewart Smith
OPAL Architect, IBM.

dcrowell77:
What do you mean by split?  Into unique repos?  packages?

@scprek @e-liner @williamspatrick - Please weigh in

stewart-ibm:
Dan <notifications@github.com> writes:
> What do you mean by split?  Into unique repos?  packages?
>
> @scprek @e-liner @williamspatrick - Please weigh in

Just into unique packages in op-build, they can each point to the
respective branch in the HB repository

-- 
Stewart Smith
OPAL Architect, IBM.

zane131:
I think I just ran into this same issue in OCC with today's op-build/master. This patch is stale: https://github.com/open-power/op-build/blob/master/openpower/package/occ/p9Patches/0001-Use-gnu89-as-C-std.patch. It fails to apply because it has already been applied. In my sandbox, I had a second patch behind it that never gets applied because of the issue with the first patch. I would expect op-build to fail. Instead, I didn't realize my patch had not been applied until I tested the code on real hardware.

PR #1359, opened 2017-08-21
Split P8 and P9 Hostboot and OCC packages

stewart-ibm:
This was causing a couple of issues:
1) increased complexity in build files
2) It breaks the pkg-version patch detection/checksum
3) Failing to apply patches to HB silently fails.

The split makes this a bit easier and requires less custom code.
We have the p8 and p9 packages provide the 'hostboot' virtual package,
allowing all the existing dependencies to work.

Fixes: https://github.com/open-power/op-build/issues/876

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1359)


stewart-ibm:
@dcrowell77 @zane131 @shenki This is what I mean when I say split HB into p8/p9 for building. This would fix the patches silently failing to apply problem as well as simplify makefiles a bit. The same thing here could be done for OCC code.

stewart-ibm:
Also fixes https://github.com/open-power/op-build/issues/1194 (I'll add that to Fixes in the commit too)

dcrowell77:
Can you elaborate a bit on how this changes the external functionality?  Would we just change to using 'hostboot-p9' as the package name for op-build commands then?

This:
> op-build hostboot-rebuild

becomes this?
> op-build hostboot-p9-rebuild

stewart-ibm:
@dcrowell77 yeah, it's exactly that. `hostboot-rebuild` would be `hostboot-p9-rebuild` and with what I just pushed for OCC to fix https://github.com/open-power/op-build/issues/1560 - same for OCC.

stewart-ibm:
I'm taking a slightly different approach to this, and I don't think this PR needs to stay open for that.

PR #1405, opened 2017-09-06
Create Combined Witherspoon Defconfig Target

wghoffa:
- All of the FW changes are in to support sequoia and
    redbud in a single config
  - This commit creates the combined/default config 'witherspoon'
  - A follow-on commit will remove the sepecifc sequoia and
    redbud configs

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1405)


stewart-ibm:

stewart-ibm:
ok to test

stewart-ibm:
We need a bit of clarification and code bump for OPAL before we merge this, but I can hit merge all at once.

dcrowell77:
Why would opal code need to change?

apopple:
> Why would opal code need to change?

Several reasons. The PLL bucket frequency needs to be in HDAT as it needs to be synced with the NVidia device driver via the device-tree. This means we need to start using the HDAT NVLink info (we don't presently). Apparently this info only exists on systems with an updated planar VPD. Therefore the OPAL code bump to match the speed change will break any systems with an old VPD, unless Hostboot is updated to interpret the Redbud vs. Sequoia from existing VPD images.

dcrowell77:
>unless Hostboot is updated to interpret the Redbud vs. Sequoia from existing VPD images
There isn't a way to do that, that's why we had them add the NV keyword.

wghoffa:
retest this please

wghoffa:
@stewart-ibm - Where are we at with this one? Are the OPAL code changes ready?

stewart-ibm:
https://github.com/open-power/op-build/pull/1443 brings in the needed OPAL changes.

We've tested on a bunch of machines with that, although machines with GPUs and updated VPD were not among them.

We also have fallback code that defaults to Sequoia in the event of not having the HDAT due to not having VPD. So, we should be good to go for this.

In fact, I'll put up a PR with both this, the skiboot bump and the removal of redbud, as it's all suitably tied in that otherwise things are going to start to look weird.

stewart-ibm:
Superseded by https://github.com/open-power/op-build/pull/1444

PR #1412, opened 2017-09-07
Update OCC Build Process for GPUs

wilbryan:
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1412)


stewart-ibm:
NAK.

We cannot have binary only stuff here, it should be *REMOVED* from the OCC repo.

wghoffa:
@wilbryan - Does Stewart's comment make sense to you? I'm not particularly sure what is binary only in this PR.

PR #1441, opened 2017-09-15
petitboot: Use linux terminal type

sammj:
Switching from vt100 to the linux termtype appears to solve the problem
of incorrectly displaying ncurses extended characters used for parts of
the window border. This error can occur, for example, by opening a SOL
session when the Petitboot UI is already drawn:

```
 Petitboot (v1.5.0-p60e3423)                          8335-GCA 2108ECA
 qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq

  System information
  System configuration
  System status log
  Language
  Rescan devices
  Retrieve config from URL
 *Exit to shell
 qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq
 Enter=accept, e=edit, n=new, x=exit, l=language, g=log, h=help
```

Switching to 'linux' correctly draws the line borders even when starting
a session late.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1441)

### DIFF
--- a/openpower/package/petitboot/petitboot-console-ui.rules
+++ b/openpower/package/petitboot/petitboot-console-ui.rules
@@ -1,5 +1,5 @@
 
 # spawn a petitboot UI on common user-visible interface devices
-SUBSYSTEM=="tty", KERNEL=="hvc*", RUN+="/usr/libexec/petitboot/pb-console --getty --detach -- -n -i 0 $name vt100"
-SUBSYSTEM=="tty", KERNEL=="tty0", RUN+="/usr/libexec/petitboot/pb-console --getty --detach -- -n -i 0 $name vt100"
-SUBSYSTEM=="tty", KERNEL=="ttyS*", RUN+="/usr/libexec/petitboot/pb-console --getty --detach -- -n -i 0 $name vt100"
+SUBSYSTEM=="tty", KERNEL=="hvc*", RUN+="/usr/libexec/petitboot/pb-console --getty --detach -- -n -i 0 $name linux"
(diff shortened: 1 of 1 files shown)


stewart-ibm:

sammj:
@jk-ozlabs please flame :) I've tested this on in a few emulators and with Putty. I'm not sure if it will make a difference or not but I've not tested Putty from a Windows box since there aren't any handy.

jk-ozlabs:
So I'm a bit sketchy on why I chose `TERM=vt100` over `TERM=linux`, but I think it was purely to be conservative with the character set. So, if it turns out this works fine in testing, then I have no issue with merging.

If the UI works with `TERM=linux` with a late-starting session, then I would assume it's using line-drawing characters rather than switching the terminal between normal and application mode. This means that the likely failure situation will be when the terminal emulator doesn't support extended character sets; I think putty has a switch to enable/disable UTF-8, perhaps that would affect this? If so, we should look into what the default setting is.

jk-ozlabs:
This is the unchanged config from a fresh install of PuTTY on Linux:

screenshot from 2017-09-27 19-25-50

- which looks fine to work with line-drawing chars. I assume the same defaults are used on Windows, but we should verify (I don't have a Windows machine to try on though...).

stewart-ibm:
Appears to be okay on what the Windows putty settings are too. Let's see if anybody complains :)

jk-ozlabs:
@stewart: you checked hyperterminal too, right? :trollface:

PR #1444, opened 2017-09-15
One witherspoon to rule them all

stewart-ibm:
Wanting to see how this all looks. I think keeping the symlink will work for now, we can update the hardcoded CI values in a week or so, then drop the symlink a week or so after that.

This would supersede https://github.com/open-power/op-build/pull/1405

Since we have a known issue in the skiboot level here for firenze, if possible I'd like to hold off pushing to master until we fix that on our monday, with me re-spinning this with that update then.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1444)


stewart-ibm:
@wghoffa I think this should be good to go know. I've kept witherspoon-sequoia.pnor just to be *sure* no CI breaks (same with keeping a copy of defconfig rather than identical).

Look good to you? (of course, here's hoping CI is good)

wghoffa:
@stewart-ibm - Looks good to me. I have retrigged the P9 HW CI, we have a few known issues floating around.

PR #1468, opened 2017-09-21
release-notes: Update to account for single witherspoon platform

sammj:
The release-notes script gets confused by the existence of both
witherspoon_sequoia_defconfig and witherspoon_defconfig. In this
scenario witherspoon_sequoia_defconfig only exists to placate the CI
process, so treat it as if it does not exist.

---
This change is [Reviewable](https://reviewable.io/reviews/open-power/op-build/1468)


sammj:
For reference: https://github.com/open-power/op-build/pull/1444

stewart-ibm:

PR #1474, opened 2017-09-21
Point to sb-signing-utils commit which implements scratch dir option

bofferdn:


stewart-ibm:

PR #1520, opened 2017-10-04
Removed the HOST_HCDB_SUPPORT directive

velozr:
Removed the HOST_HCDB_SUPPORT directive from the config file. It
will now be controlled from the hwas/HBconfig file.


stewart-ibm:

dcrowell77:
For historical purposes some more info - This flag controls 2 things that would be visible to the customer.
1) Do we detect hardware changes and clear gard records?
2) Do we detect new hardware and run expanded memory patterns in memory diagnostics?

Prior to this change, the answer to both of those questions was no, now it is (correctly) yes.

PR #1540, opened 2017-10-10
Secure Boot: Point op-build at sb-signing-utils archive support commit

bofferdn:


stewart-ibm:

PR #1541, opened 2017-10-11
Fix OCC Make process when using GPU Binary

wghoffa:


wghoffa:
Only failure is an internal OpenBMC error. Merging.

PR #1563, opened 2017-10-18
Modify core and thread level events for POWER9 DTS files

drajarshi:
Segregate DD1 and DD2 core and thread level events.

Signed-off-by: Raj <drajarshi@in.ibm.com>


stewart-ibm:

PR #1566, opened 2017-10-18
Update ppe42-gcc to latest version. 10/2017

dgilbert999:


stewart-ibm:

stewart-ibm:
ok to test

stewart-ibm:
issues with this PR:
- no Signed-off-by
- keeps old entry around, which could be just found by querying git history
- doesn't include `git shortlog` output in commit (which is not needed, but nice)

I opened https://github.com/open-power/op-build/pull/1569 which does the bump that way, so I'll merge it instead.

PR #1580, opened 2017-10-23
kernel: Move to Linux 4.13.8-openpower1

sammj:
Move to v4.13.8 and include another patch from next:
45348de "scsi: aacraid: Fix controller initialization failure"

This fixes a problem with the PM8069 controller used on the p9dsu
platform causing the controller to crash after a reset.


stewart-ibm:

PR #1582, opened 2017-10-24
Bump sb-signing-utils version to v0.3

hellerda:
Bump sb-signing-utils to tagged version v0.3


hellerda:
Hi Nick.  Everything tagged v0.3 is tested in development mode and production mode on a current version of op-build.  There have been a few commits since I last demoed on 10/13, but they are all trivial patches, or patches I've been running with in my test env for a while.  Let me know if any issues or questions.  Thx -Dave

PR #1584, opened 2017-10-25
kernel: Move to Linux 4.13.8-openpower2

sammj:
This replaces our existing non-upstream xhci fix with another
non-upstream xhci fix. It goes back to the original
usb_hcd_pci_shutdown() handler and instead performs a hardware reset of
the controller in xhci_shutdown().

Discussion of this change can be found here:
https://marc.info/?l=linux-usb&m=150853502520762&w=2

This fixes a particular issue on p9dsu, and after some testing does not
appear to regress on other platforms.


stewart-ibm:

stewart-ibm:
ok to test

PR #1610, opened 2017-11-07
Buildroot 2017.08.1 and kernel 4.13.11

shenki:


shenki:
Supersedes https://github.com/open-power/op-build/pull/1605

PR #1623, opened 2017-11-15
Kernel 4.14

shenki:


shenki:
It appears the IBM OP CI machines are unable to switch paths for the git submodule, so this change can't be tested by them :(

shenki:
Tested by hand on Romulus. Should be good to go.

PR #1624, opened 2017-11-15
Buildroot 2017.11

shenki:
For testing only. Do not merge yet.


shenki:
It appears the IBM OP CI machines are unable to switch paths for the git submodule, so this change can't be tested by them :(

shenki:
We have now moved to Buildroot 2017.11

PR #1656, opened 2017-11-27
kernel: Move to Linux v4.14.2-openpower1

shenki:


stewart-ibm:
CI failures are false negatives, ignoring.

PR #1673, opened 2017-12-01
Buildroot 2017.11 and Linux 4.14.3

shenki:


stewart-ibm:
Ahhh Jankins - where all tests passing is actually a failure. Good times.

PR #1707, opened 2017-12-14
Romulus update the latest XML

MengzeLiao:
Update the romulus xml to the latest version which fixed the FSI problem on the slave proc


stewart-ibm:

PR #1831, opened 2018-02-05
Buildroot 2017.11.2 and Kernel 4.15.1-openpower-1

shenki:


stewart-ibm:
Unless there's any strong objection, I'd prefer to take this *after* we tag the op-build v1.21 release.

I'm also going a bit either way on if we should stick with the LTS v4.14 kernel for v2.00 or if just shipping "latest stable" is going to be okay...

stewart-ibm:
@shenki Considering we're really close to tagging v1.21 and I've moved things over to the release branch rather than master, should we go for 4.15.5 (latest stable)

shenki:
Can we merge this and then I'll bump to the latest kernel on Monday?

stewart-ibm:
@shenki yeah. I'll resolve the merge conflicts here along with the https://github.com/open-power/op-build/pull/1900 and https://github.com/open-power/op-build/pull/1882 changes (ipr fast reset and enabling powersave)

shenki:
Replaced with https://github.com/open-power/op-build/pull/1908

PR #1922, opened 2018-02-28
hostboot & hostboot-binaries bump, disable stop4 stop5 on witherspoon

stewart-ibm:
Supersedes https://github.com/open-power/op-build/pull/1916


stewart-ibm:
retest this please

PR #1931, opened 2018-03-01
Add support for Stonewither to open-power

LowellDennis:
The Stonewither board is project undertaken by the Dell ESI team and the IBM OpenPower team to create an adaptation of IBMs Witherspoon board that fits in Dells G5 chassis.


stewart-ibm:

stewart-ibm:
ok to test

openpower/patches/stonewither-patches/linux/0001-power-fixed2.patch
```diff
@@ -0,0 +1,17 @@
+--- a/drivers/net/ethernet/intel/fm10k/fm10k_pci.c	2018-01-10 15:27:42.437514511 -0600
```
stewart-ibm: We like kernel patches to go into upstream kernels and we pick them up through the stable tree (we consider any local patches we're carrying to be a bug :)

Has this one been submitted to upstream stable?

If we need something before the next stable kernel is released, we'll pick it up, but globally for op-build so that we don't end up with many kernel trees to maintain.

openpower/configs/stonewither_defconfig
```diff
@@ -0,0 +1,67 @@
...
+BR2_ROOTFS_DEVICE_TABLE="../openpower/device_table.txt"
+BR2_TARGET_GENERIC_GETTY_PORT="hvc0"
+BR2_ENABLE_LOCALE_WHITELIST="C de en es fr it ja ko pt_BR ru zh_CN zh_TW"
+BR2_GENERATE_LOCALE="en_US.UTF-8 de_DE.UTF-8 es_ES.UTF-8 fr_FR.UTF-8 it_IT.UTF-8 ja_JP.UTF-8 ko_KR.UTF-8 pt_BR.UTF-8 ru_RU.UTF-8 zh_CN.UTF-8 zh_TW.UTF-8"
+BR2_SYSTEM_ENABLE_NLS=y
+BR2_ROOTFS_OVERLAY="../openpower/overlay"
+BR2_ROOTFS_POST_BUILD_SCRIPT="../openpower/scripts/fixup-target-var ../openpower/scripts/firmware-whitelist"
+BR2_LINUX_KERNEL=y
+BR2_LINUX_KERNEL_CUSTOM_VERSION=y
+BR2_LINUX_KERNEL_CUSTOM_VERSION_VALUE="4.14.16"
```
stewart-ibm: We're up to 4.15.6 now, and your config file was probably created before that. It should be okay to just bump the version here (diff with witherspoon_defconfig).

openpower/configs/stonewither_defconfig
```diff
@@ -0,0 +1,67 @@
+BR2_powerpc64le=y
+BR2_powerpc_power8=y
+BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_OP_BUILD_PATH)/patches/stonewither-patches"
```
stewart-ibm: Should this currently just point to witherspoon-patches? I can imagine the same issues we sometimes patch around on witherspoon would apply to stonewither?
(I'm thinking of our current stop4/stop5 disabling on witherspoon)

output/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
stewart-ibm: This file shouldn't be removed.

stewart-ibm:
Superseded by https://github.com/open-power/op-build/pull/1942

PR #1965, opened 2018-03-14
kernel: Move to Linux 4.15.9-openpower1

shenki:


stewart-ibm:
(I reran the failing p9 ipl test and it passed. I'm guessing this is one of the lovely sporadic problems we have with ancient kernels like the one involved in the ipl test)

PR #1966, opened 2018-03-14
Optimize Busybox configuration

stewart-ibm:
There's a number of things we do *not* need in the limited
petitboot environment, so we should exclude them.

We also enable the taskset utility as we can use it in op-test
to help checking cpu stop states, and that's reason enough.

Before 989872 (358760 xz compressed)
After  918816 (339628 xz compressed)

A saving of 19132 bytes!


shenki:
I agree with all of the things you've removed are okay to remove. Is it worth the maintenance burden carrying this change given the small payoff?

sammj:
So after this morning's conversation about passwords - I'm pretty sure we're going to want `CONFIG_FEATURE_SHADOWPASSWDS`, `CONFIG_PASSWD`, etc etc :smile:

stewart-ibm:
> I agree with all of the things you've removed are okay to remove. Is
> it worth the maintenance burden carrying this change given the small
> payoff?

Yeah, it probably is. Adding taskset is probably the only thing I really
want (purely for testing purposes)

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
Sam Mendoza-Jonas <notifications@github.com> writes:
> So after this morning's conversation about passwords - I'm pretty sure
> we're going to want `CONFIG_FEATURE_SHADOWPASSWDS`, `CONFIG_PASSWD`,
> etc etc :smile:

Do we? Or do we want a pam module that does the right thing? Either way,
I think I should ditch this PR.

-- 
Stewart Smith
OPAL Architect, IBM.

PR #1967, opened 2018-03-14
Docs: Document supported hardware

shenki:
This list is taken from the current kernel configuration. The intent is
it will be updated as requests come in to add new hardware.


shenki:
I'm open to suggestions as to where the link from the toc goes

PR #1972, opened 2018-03-16
Patch set for VESNIN server support

artemsen:
I can integrate these patches into OCC/Hostboot (P8 branch) with a configurable build via the c-preprocessor (defines), but I doubt that this work will be useful given that the P8 branches are almost not developing and these patches apply only to our hardware implementation.

### DIFF
--- a/openpower/patches/vesnin-patches/hostboot/hostboot-0001-Increase-default-watchdog-timeout-to-300-seconds.patch
+++ b/openpower/patches/vesnin-patches/hostboot/hostboot-0001-Increase-default-watchdog-timeout-to-300-seconds.patch
@@ -1,0 +1,29 @@
+From a4259b9799ebb0f31223d52def871f4851217315 Mon Sep 17 00:00:00 2001
+From: Artem Senichev <a.senichev@yadro.com>
+Date: Tue, 13 Mar 2018 14:50:56 +0300
+Subject: [PATCH] Increase default watchdog timeout to 300 seconds
+
+We need more than 2 minutes to go through some IPL, for instance
+step "13.10 mss_draminit_trainadv" tooks 215 seconds to initialize 8 TiB RAM.
+
+Signed-off-by: Artem Senichev <a.senichev@yadro.com>
+---
+ src/include/usr/ipmi/ipmiwatchdog.H | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/include/usr/ipmi/ipmiwatchdog.H b/src/include/usr/ipmi/ipmiwatchdog.H
+index d730e6bc4..37f73624b 100644
+--- a/src/include/usr/ipmi/ipmiwatchdog.H
++++ b/src/include/usr/ipmi/ipmiwatchdog.H
+@@ -58,7 +58,7 @@ namespace IPMIWATCHDOG
+ #ifdef CONFIG_BMC_IPMI_LONG_WATCHDOG
+ const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 2400;
+ #else
+-const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 120;
++const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 300;
+ #endif
+ 
+ /**
+-- 
+2.14.1
+
--- a/openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
+++ b/openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
@@ -3,6 +3,12 @@ From: Artem Senichev <a.senichev@yadro.com>
 Date: Tue, 13 Mar 2018 15:03:41 +0300
 Subject: [PATCH] Fix POS and SPD-VPD_REC_NUM
 
+Hank Chang: When bringup this 4Socket 128DIMMs system, we hit the following
+VPD/SPD issue and suspected it might be related to the VPD_REC_NUM numbering
+algorithm issue then caused the SPD cache overlap for the 4 sockets case.
+So we simply tried to jump the $aff_pos offset from 16 to 64 for each Proc and
+the issue was fixed/gone.
+
 Originally created by MSI (S188)
 
 Signed-off-by: Artem Senichev <a.senichev@yadro.com>
--- a/openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
+++ b/openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
@@ -1,0 +1,28 @@
+From 1dcf087af57a93ec7f24aa8be73e7ff0ffad22ca Mon Sep 17 00:00:00 2001
+From: Artem Senichev <a.senichev@yadro.com>
+Date: Tue, 13 Mar 2018 15:03:41 +0300
+Subject: [PATCH] Fix POS and SPD-VPD_REC_NUM
+
+Originally created by MSI (S188)
+
+Signed-off-by: Artem Senichev <a.senichev@yadro.com>
+---
+ src/usr/targeting/common/Targets.pm | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/usr/targeting/common/Targets.pm b/src/usr/targeting/common/Targets.pm
+index 35f08428c..4eaba16e7 100644
+--- a/src/usr/targeting/common/Targets.pm
++++ b/src/usr/targeting/common/Targets.pm
+@@ -744,7 +744,7 @@ sub processMcs
+                         $self->setAttribute($dimm,"MBA_PORT",$port_num);
+                         $self->setAttribute($dimm,"MBA_DIMM",$dimm_num);
+ 
+-                        my $aff_pos=16*$proc+$mcs*$self->{MAX_MCS}+4*$mba+
++                        my $aff_pos=64*$proc+$mcs*$self->{MAX_MCS}+4*$mba+
+                                     2*$port_num+$dimm_num;
+                         $self->setAttribute($dimm, "AFFINITY_PATH",
+                             $parent_affinity
+-- 
+2.14.1
+
--- a/openpower/patches/vesnin-patches/hostboot/hostboot-0003-Add-present-bits-info-based-on-pos.patch
+++ b/openpower/patches/vesnin-patches/hostboot/hostboot-0003-Add-present-bits-info-based-on-pos.patch
@@ -1,0 +1,97 @@
+From e89b453239c0b087e5f61aefa59d29a1ae636e08 Mon Sep 17 00:00:00 2001
+From: Artem Senichev <a.senichev@yadro.com>
+Date: Tue, 13 Mar 2018 15:14:49 +0300
+Subject: [PATCH] Add present bits info based on pos
+
+Originally created by MSI (S188)
+
+Signed-off-by: Artem Senichev <a.senichev@yadro.com>
+---
+ src/usr/hwas/hostbootIstep.C | 52 ++++++++++++++++++++++++++++++++++++++++++++
+ 1 file changed, 52 insertions(+)
+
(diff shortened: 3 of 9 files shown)


stewart-ibm:

stewart-ibm:
ok to test

stewart-ibm:
@sannerd @dcrowell77 what do you both think of how we should do this? In skiboot/linux land we're all about making upstream work for everyone, but I know that P8 hostboot is pretty low on the priority list there.

openpower/patches/vesnin-patches/hostboot/hostboot-0001-Increase-default-watchdog-timeout-to-300-seconds.patch
```diff
@@ -0,0 +1,29 @@
...
+diff --git a/src/include/usr/ipmi/ipmiwatchdog.H b/src/include/usr/ipmi/ipmiwatchdog.H
+index d730e6bc4..37f73624b 100644
+--- a/src/include/usr/ipmi/ipmiwatchdog.H
++++ b/src/include/usr/ipmi/ipmiwatchdog.H
+@@ -58,7 +58,7 @@ namespace IPMIWATCHDOG
+ #ifdef CONFIG_BMC_IPMI_LONG_WATCHDOG
+ const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 2400;
+ #else
+-const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 120;
++const uint16_t  DEFAULT_WATCHDOG_COUNTDOWN = 300;
```
dcrowell77: Curious why you needed to do this.  Was there a specific istep that was taking more than 2 minutes to complete?

openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/src/usr/targeting/common/Targets.pm b/src/usr/targeting/common/Targets.pm
+index 35f08428c..4eaba16e7 100644
+--- a/src/usr/targeting/common/Targets.pm
++++ b/src/usr/targeting/common/Targets.pm
+@@ -744,7 +744,7 @@ sub processMcs
+                         $self->setAttribute($dimm,"MBA_PORT",$port_num);
+                         $self->setAttribute($dimm,"MBA_DIMM",$dimm_num);
+ 
+-                        my $aff_pos=16*$proc+$mcs*$self->{MAX_MCS}+4*$mba+
++                        my $aff_pos=64*$proc+$mcs*$self->{MAX_MCS}+4*$mba+
```
dcrowell77: Was there a reason you needed numbers this high?  Physically there can only be 16 dimms behind each processor.

openpower/patches/vesnin-patches/hostboot/hostboot-0003-Add-present-bits-info-based-on-pos.patch
```diff
@@ -0,0 +1,97 @@
...
+         }
+         TARGETING::TYPE l_type = target->getAttr<TARGETING::ATTR_TYPE>();
+         TARGETING::ATTR_POSITION_type l_pos = 0;
++        TARGETING::ATTR_HUID_type l_huid = target->getAttr<TARGETING::ATTR_HUID>(); //for showing HUID
++
+         if( target->tryGetAttr<TARGETING::ATTR_POSITION>(l_pos) )
+         {
+             l_presData[l_type] |= (0x8000000000000000 >> l_pos);
++
++            if (l_type == 0x03 ) //if the type is DIMM
```
dcrowell77: There are constants/enums for the types versus using a literal

openpower/patches/vesnin-patches/hostboot/hostboot-0005-Fill-empty-sensor-id-to-reserved-id-0xFF.patch
```diff
@@ -0,0 +1,73 @@
...
+         else
+         {
+-            TRACFCOMP(g_trac_ipmi,"We were not able to find a sensor number in"
++            TRACFCOMP(g_trac_ipmi,"Found a reserved sensor number (0xFF) in"
+                       " the IPMI_SENSORS attribute for sensor_name=0x%x"
+                       "for target with huid=0x%x, skipping call to "
+                       "sendSetSensorReading()",
+                     iv_name, TARGETING::get_huid( iv_target ));
+-
+-            assert(false);
```
dcrowell77: So if you request an unknown sensor it just does nothing?

openpower/patches/vesnin-patches/hostboot/hostboot-0005-Fill-empty-sensor-id-to-reserved-id-0xFF.patch
```diff
@@ -0,0 +1,73 @@
...
+-
+         }
+ 
+         return l_err;
+diff --git a/src/usr/targeting/common/processMrw.pl b/src/usr/targeting/common/processMrw.pl
+index 8bfa7bddd..45d250872 100644
+--- a/src/usr/targeting/common/processMrw.pl
++++ b/src/usr/targeting/common/processMrw.pl
```
dcrowell77: Can you summarize what you're trying to do here?

openpower/patches/vesnin-patches/hostboot/hostboot-0007-Set-IBM-instead-of-AMI-for-eSEL.patch
```diff
@@ -0,0 +1,30 @@
...
+diff --git a/src/include/usr/ipmi/ipmiif.H b/src/include/usr/ipmi/ipmiif.H
+index 44283cc0b..0f1309e31 100644
+--- a/src/include/usr/ipmi/ipmiif.H
++++ b/src/include/usr/ipmi/ipmiif.H
+@@ -258,7 +258,7 @@ namespace IPMI
+ 
+     //AMI-specific storage messages
+     inline const command_t partial_add_esel(void)
+-    { return std::make_pair(NETFUN_AMI, 0xf0); }
++    { return std::make_pair(NETFUN_IBM, 0xf0); }
```
dcrowell77: FYI - we made this change in our P9 codebase too, but we were wary to make it in the legacy code.

dcrowell77:
I've got no inherent problem with merging in some patches, but I do see a lot of very specific changes in here.  We'd need some CONFIG flags around stuff.

I actually like some of the display changes and I'd entertain the idea of pulling them into master as well since the same code exists there.  We would need to plumb up some generic ways to break things down though.  I do know that other folks have made some similar tweaks in the past.

openpower/patches/vesnin-patches/hostboot/hostboot-0001-Increase-default-watchdog-timeout-to-300-seconds.patch
artemsen: At least step "13.10 mss_draminit_trainadv" tooks 215 seconds to initialize 8 TiB RAM.

openpower/patches/vesnin-patches/hostboot/hostboot-0007-Set-IBM-instead-of-AMI-for-eSEL.patch
stewart-ibm: It should possibly be an attribute that's settable in the machine XML... or something detected at runtime. Supermicro has the same issue for their BMC. I think they solve it by patching hostboot in a similar way (in skiboot, we currently have a per-platform setting).

I'm having some discussions with AMI and OpenBMC about this too, as there are systems that can have an AMI or OpenBMC, and it'd be better if we didn't have to have different host firmware to support that change.

openpower/patches/vesnin-patches/hostboot/hostboot-0005-Fill-empty-sensor-id-to-reserved-id-0xFF.patch
HankChang: Hi @artemsen, I think you could try to remove this patch once all the set sensor related IDs are assigned properly. The patch was just a quick workaround for speeding up system bringup w/o code assert.

openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
HankChang: When bringup this 4Socket 128DIMMs system, we hit the following VPD/SPD issue and suspected it might be related to the VPD_REC_NUM numbering algorithm issue then caused the SPD cache overlap for the 4 sockets case. (attached the default output VPD_REC_NUM_ORG.txt, you would see some duplicated numbering starting from 3rd CPU's DIMMs like VPD_REC_NUM=32..47)
[VPD_REC_NUM_ORG.txt](https://github.com/open-power/op-build/files/1858672/VPD_REC_NUM_ORG.txt)
```

 16.23577|================================================
 16.23577|Error reported by vpd (0x1D00)
 16.24571|  Invalid Basic Memory Type
 16.24572|  ModuleId   0x61 VPD::VPD_SPD_GET_KEYWORD_VALUE
 16.24572|  ReasonCode 0x1d07 VPD::VPD_INVALID_BASIC_MEMORY_TYPE
 16.24572|  UserData1  Basic Memory Type (Byte 2) : 0x00000000000000ff
 16.24573|  UserData2  Keyword Requested : 0x0000000000000025
```
 
So we simply tried to jump the $aff_pos offset from 16 to 64 for each Proc and the issue was fixed/gone. 
[VPD_REC_NUM_MOD.txt](https://github.com/open-power/op-build/files/1858673/VPD_REC_NUM_MOD.txt)

openpower/patches/vesnin-patches/hostboot/hostboot-0007-Set-IBM-instead-of-AMI-for-eSEL.patch
dcrowell77: Remember that this is the P8 tree.  We already made this change in our P9 code as we got everything to agree.

openpower/patches/vesnin-patches/hostboot/hostboot-0005-Fill-empty-sensor-id-to-reserved-id-0xFF.patch
HankChang: Discussed with MSI team, seems like this specific processMrw.pl patch could help them achieve the following goals:
1. Auto set all empty sensors with reserved ID - 0xFF instead of being 0x0 in the mrw process (would cause some duplicate sensor@0 in skiboot)

2. In this way, they could simply save 128 empty DIMM_Temp* IDs to become all 0xFF, and OCC would still report all the non-zero ID's DIMM_Temp* reading in the poll response.  (https://github.com/open-power/occ/blob/master_p8/src/occ/cmdh/cmdh_fsp_cmds.c#L392)

And then, apply another "occ-0001-Add-DIMM-temperature-sensors.patch" (in this pull request) can customize/override these 0xFF temp_sid again based on the different CPU/Cen/DIMM index which can be decoded by OpenBMC accordingly.

3. Just simply prints the "N/A" (0xFF) string in the *.rpt file for reviewing all the undefined/reserved sensor targets.

openpower/patches/vesnin-patches/hostboot/hostboot-0002-Fix-POS-and-SPD-VPD_REC_NUM.patch
HankChang: BTW, we've also checked a booted PNOR dump with 128DIMMs installed, seems like all the 128 DIMM SPD cache regions can be updated properly within the DJVPD partition range. (0x91000-0xD9000)
[DJVPD_Dump.bin.zip](https://github.com/open-power/op-build/files/1864297/DJVPD_Dump.bin.zip)

artemsen:
I have added description to patches from this discussion written by @HankChang.
@stewart-ibm, would you disable the p9 test for the VESNIN platform?
@dcrowell77, do you have any more questions about these patches?

dcrowell77:
Yeah, merging this is fine.  We can sort out patches separately at a slower pace

artemsen:
Thanks! It was the last diff between our branches.

stewart-ibm:
Excellent! I like it when we can get everything upstream!

PR #2067, opened 2018-04-26
Use the OP920 branches for Hostboot and HCODE

stewart-ibm:
It turns out that OCC and SBE are the same in the current op920 branch, so they do *not* get the same treatment. We can add it if they do.

This series attempts to get us to the point where op-build master can be used to get something that everyone agrees is a good base for POWER9 systems as we head towards op-build v2.0.

As such, the aim is to have these stable branches (named op920 for reasons that aren't terribly interesting) for some firmware components be the ones that are built by default, while at the same time, allowing a config option that will lead you down the joyful path of running whatever the latest code is.

To achieve this and not get terribly complex, the first step is splitting off the POWER8 branches for OCC and Hostboot into separate packages: occ-p8 and hostboot-p8. Since these need different sets of patches and that's caused problems in the past, this is the "easy" way to solve that before we start adding in the (small) complexity of having selectable versions.

To build the 'master' versions, you'd do:
```
BR2_HCODE_LATEST_VERSION=y
BR2_HOSTBOOT_LATEST_VERSION=y
```
in your config rather than taking the default option of the stable OP920 branches. I'm open to the idea of having a single option that enables all of these kind of options, perhaps even a separate defconfig?

Additionally, a few tiny cleanups of how config options are presented in menus are included, if only to help me work out that I think I got things right.

There are two temporary patches in this PR that point the OP920 branches to the ibm-op-release project. If we go ahead and merge this, then these patches should go away and the stable branches be mirrored to open-power rather than that separate project.

The split in versions here is likely a temporary thing as we work out the best ways to manage the situation we're in and the development process for these components.


stewart-ibm:
Fixed the `BR2_OCC_GPU_BIN_BUILD=y` in the latest set of patches, as well as adding a p9dus and witherspoon defconfig that selects the master versions

stewart-ibm:
Incorporates version bumps from:
- https://github.com/open-power/op-build/pull/2072
- https://github.com/open-power/op-build/pull/2070
- and op920 branches from https://github.com/ibm-op-release/op-build/pull/216

dcrowell77:
This is an 'interesting' set of changes.

> have these stable branches (named op920 for reasons that aren't terribly interesting) for some firmware components be the ones that are built by default

To me that just obfuscates what is happening.  Why would the primary op-build repo point to a fork as the default option?

stewart-ibm:
> This is an 'interesting' set of changes.
>
>> have these stable branches (named op920 for reasons that aren't terribly interesting) for some firmware components be the ones that are built by default
>
> To me that just obfuscates what is happening.  Why would the primary
> op-build repo point to a fork as the default option?

So that the testing and focus coming up to op-build v2.0 is on what
we'll end up tagging and releasing as "use this as the minimum starting
point for power9 systems"

... and maybe, just *maybe* we'll be able to get a single combination of all
the components where everything works correctly :/


-- 
Stewart Smith
OPAL Architect, IBM.

PR #2081, opened 2018-05-02
Split OCC and Hostboot p8 and p9, add op920 branches as stable branches for PFW components to support getting to a op-build v2.0 release that's suitable for GA

stewart-ibm:
Supersedes https://github.com/open-power/op-build/pull/2067 as it's no longer an RFC.

This series gets us to the point where op-build master can be used to get something that everyone agrees is a good base for POWER9 systems as we head towards op-build v2.0.

These stable branches (named op920 for reasons that aren't terribly interesting) for some firmware components be the ones that are built by default, while at the same time, allowing a config option that will lead you down the joyful path of running whatever the latest code is.

To achieve this and not get terribly complex, the first step is splitting off the POWER8 branches for OCC and Hostboot into separate packages: occ-p8 and hostboot-p8. Since these need different sets of patches and that's caused problems in the past, this is the "easy" way to solve that before we start adding in the (small) complexity of having selectable versions.

To build the 'master' versions, you'd do:
```
BR2_HOSTBOOT_LATEST_VERSION=y
BR2_HOSTBOOT_BINARIES_LATEST_VERSION=y
BR2_HCODE_LATEST_VERSION=y
BR2_OCC_LATEST_VERSION=y
BR2_SBE_LATEST_VERSION=y
```
in your config rather than taking the default option of the stable OP920 branches. There are also `p9dsu_dev_defconfig` and `witherspoon_dev_defconfig` to make it easy to do.

I do not see this as a long term solution in the state it is now, we'll want to discuss how we deal with this going forward, and I aim to do so in person over the next couple of weeks, as well as set things up into the future. This series of patches is solving our *current* problem, not all our problems forever :)

Additionally, a few tiny cleanups of how config options are presented in menus are included, if only to help me work out that I think I got things right.

I've pushed the `release-op920` branches from ibm-op-release/ projects to the main `open-power` ones here, so we are going to be picking up the correct versions of things and have the git sha1s exist in upstream repositories.


dcrowell77:
I still have the same concerns as I had on https://github.com/open-power/op-build/pull/2067 , but I hear there is going to be a pow-wow next week to discuss this so I'll just let it be.

PR #2087, opened 2018-05-03
Default Hostboot to Latest Code

wghoffa:


mzipse:
@wghoffa , why are you doing this?   We are communicating to the OpenPower community that the default is the stable branch (ie OP920)......which is what I've been told they want.  Plus, if we are going to change the default for Hostboot, we need to change it for the other repos too to be consistent.

stewart-ibm:
We get this functionality as part of https://github.com/open-power/op-build/pull/2119 - which is post op-build v2.0

PR #2099, opened 2018-05-08
linux: Enable ipmi watchdog support

wak-google:
We need this support so that the skiroot linux kernel can handle the
watchdog after it is handed off by skiboot. This way we can keep the
watchdog enabled throughout the entire skiboot handoff to linux instead
of disabling it prior to executing the linux kernel. This allows us to
increase our watchdog coverage during boot.

It also gives us the benefit that we can now use the ipmi watchdog
through /dev/watchdog in the skiroot environment.


jk-ozlabs:
OK, so it does disable before kexec, that's great. All fine by me then.

wak-google:
Anyone else I need to sign off on this?

shenki:
LGTM. @stewart-ibm can we please merge this?

stewart-ibm:
"William A. Kennington III" <notifications@github.com> writes:
> Anyone else I need to sign off on this?

Let me run it through a few tests and I'll be okay to merge it.

I trust the CI about as far as I can throw a gorilla, so I'll run actual
tests :)
-- 
Stewart Smith
OPAL Architect, IBM.

PR #2119, opened 2018-05-17
sbe, hostboot, hostboot-binaries, hcode bumps, no longer do OP920 branches

stewart-ibm:
This gets us back to building the master branches for the PFW bits.


stewart-ibm:
Breadcrumbs for @ravimed1 @dcrowell77 @sannerd @wghoffa @pragupta and whoever else I was discussing this with and have now forgotten as it's late and I'm in LAX and thus everything is terrible.

PR #2150, opened 2018-05-30
petitboot: Update to v1.8.0 and switch to tarballs

sammj:
Petitboot now distributes tarballs so use and verify them; this avoids
having to perform extra autoconf magic before configuring Petitboot.


stewart-ibm:

PR #2151, opened 2018-05-30
openpower: Add msglog tool to skiroot

shenki:
This adds a one line shell script that can be used to display the OPAL
msglog without having to type out the full path.

### DIFF
--- a/openpower/overlay/bin/msglog
+++ b/openpower/overlay/bin/msglog
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec cat /sys/firmware/opal/msglog


openpower/overlay/bin/msglog
```diff
@@ -0,0 +1,2 @@
+#/bin/sh
```
jk-ozlabs: You're missing the `!`

openpower/overlay/bin/msglog
```diff
@@ -0,0 +1,2 @@
...
+cat '/sys/firmware/opal/msglog'
```
jk-ozlabs: `exec cat` ?

shenki:
Updated following @jk-ozlabs 's comments

PR #2152, opened 2018-05-30
Rename Vesnin patch directories

artemsen:
Commit 03671c split P8 and P9 hostboot/OCC packages, so we need to
rename patch directories according to the new package names:
hostboot -> hostboot-p8
occ -> occ-p8


stewart-ibm:
ok to test

PR #2181, opened 2018-06-11
HDAT pnor partition population with hdat binary data.

nagendraIBM:
1> Check the presence of hdat binary generation script.
2> Run the script to generate hdat binary file.
3> Populate the HDAT pnor partition with hdat binary data.


stewart-ibm:

PR #2194, opened 2018-06-21
Modify NX unit name in POWER9 DTS file

kgdeepak:
Modify NX unit name in POWER9 DTS file


stewart-ibm:

PR #2222, opened 2018-07-11
2.0.y pfw bumps 20180711

stewart-ibm:


stewart-ibm:
retest this please

stewart-ibm:
I have to drop the HCODE bump as it causes hard lockups.

PR #2245, opened 2018-07-25
ci: Use uname rather than arch

sammj:
Also fixes the failure message.


stewart-ibm:

PR #2252, opened 2018-07-30
kernel: Move to Linux v4.17.11-openpower1

shenki:
PowerPC fixes:

 KVM: PPC: Check if IOMMU page is contained in the pinned physical page
 powerpc/powernv: Fix save/restore of SPRG3 on entry/exit from stop (idle)

Driver fixes:

 net/mlx5: Adjust clock overflow work period
 net/mlx5e: Fix quota counting in aRFS expire flow
 net/mlx5e: Don't allow aRFS for encapsulated packets
 net/mlx5e: Refine ets validation function
 net/mlx5e: Only allow offloading decap egress (egdev) flows
 net/mlx5e: Add ingress/egress indication for offloaded TC flows
 net/mlx5: E-Switch, UBSAN fix undefined behavior in mlx5_eswitch_mode
 net/mlx4_core: Save the qpn from the input modifier in RST2INIT wrapper
 net/mlx4_en: Don't reuse RX page when XDP is set
 tg3: Add higher cpu clock for 5762.


stewart-ibm:
retest this please

PR #2265, opened 2018-08-06
Improvements to the CI build scripts

sammj:
Some small enhancements to the CI build scripts, mainly motivated by trying to set up a Jenkins-in-docker server behind a proxy:
- Use `--network=host`
- Allow a custom mount point (so we can..)
- Specify custom locations for the download and ccache directories.


stewart-ibm:

PR #2290, opened 2018-08-20
kernel: Move to Linux 4.18.3-openpower1

shenki:
Update to 4.18, and re-generate the defconfig. In addition, make the
following changes:

 - Enable EXPERT, so we can disable some things
 - Disable non-powerpc BPF decoders
 - Disable TASKSTATS
 - Disable unused syscalls
 - Set more things to be modules
 - Turn off unused network vendors

Note that CONFIG_I2C_CHARDEV=y as the module is not loaded
automatically, and without this i2cget etc. will fail.

Buildroot is updated to include support for v4.18.


shenki:
Tested with op-test in qemu, and on Romulus

PR #2310, opened 2018-08-28
Updates to CI scripts

sammj:
Tweak where `build-all-defconfigs.sh` cleans up its output directory to make post-build processing and/or debugging easier, and add an option to generate the `legal-info` target during the build.


stewart-ibm:
(if your wondering why my CI failed there, it's because I've been applying some hacky patches to the CI scripts that I've been meaning to clean up and send upstream). I should rebase my patches on top of yours, as part of it goes down a similar route.

PR #2319, opened 2018-08-30
CI: Speed up builds

stewart-ibm:
- support building and caching of SDK
- use eatmydata
- build more than just legal-info, for the reasons.

### DIFF
--- a/ci/build-all-defconfigs.sh
+++ b/ci/build-all-defconfigs.sh
@@ -8,18 +8,24 @@ CONFIGTAG="_defconfig"
 
 DEFCONFIGS=();
 
-while getopts "o:p:r" opt; do
+SDK_DIR=""
+
+while getopts "o:p:rs:" opt; do
   case $opt in
     o)
       echo "Output directory: $OPTARG"
       OUTDIR="$OPTARG"
+      ;;
+    s)
+      echo "SDK is in: $OPTARG"
+      SDK_DIR=$OPTARG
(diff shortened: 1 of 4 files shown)


ci/build-all-defconfigs.sh
```diff
@@ -62,25 +62,20 @@ if [ -n "$DL_DIR" ]; then
...
+        op-build O=$O
         r=$?
-
-        if [ ${BUILD_INFO} = 1 ] && [ $r = 0 ]; then
-                op-build legal-info
-                mv output/legal-info ${OUTDIR}/$i-legal-info
-        fi
-
-        mkdir ${OUTDIR}/$i-images
-        mv output/images/* ${OUTDIR}/$i-images/
```
sammj: I think we lose something here. Instead of moving the output images of each platform to
`${OUTDIR}/platform_name-images/` we leave everything in the `${OUTDIR}` - which is fine for building but not if we want to grab the images.
We also don't `rm -rf` the output directory (now `${OUTDIR}`) so each platform will build on top of itself I think.

stewart-ibm: yeah, fair enough on cleaning things out. I'll fix that out.

We should be passing through `pwd` to the container, so we'll end up with a bunch of output dirs in `pwd` with all the build artifacts. This is either a feature or a bug I guess, depending on how much disk space you have.

PR #2342, opened 2018-09-13
Add bsdmainutils to Ubuntu docker

artemsen:
The docker image of Ubuntu 16.04 doesn't have 'colrm' utility which is
used in Perl script 'addimgid' (hostboot build tools) to construct an ID string
of hostboot image:
https://github.com/open-power/hostboot/blob/master/src/build/tools/addimgid#L39
As result, variable hbi_ImageId in hostboot is allways empty.
Installing bsdmainutils package resolves this problem.


stewart-ibm:
ok to test

stewart-ibm:
Ignore the Travis CI failure, something new has started breaking there when building the Latex documentation, it's not your patch

PR #2346, opened 2018-09-17
Enable parallel make in SBE

Shakeebbk:
Can be merged once da99f39710df224314573b411f0fd5ca8743c743 is in.


stewart-ibm:
So, when I tried with both patches applied, a SBE build (the rebuild part of `op-build sbe-dirclean && op-build sbe-rebuild` ) seemed a lot quicker:
```
real    1m55.991s
user    23m34.785s
sys     1m43.853s
```

Which is fantastic! Thanks a lot for this!

stewart-ibm:
I've made the equivalent change in 82f33c72e509 as your commit didn't have `Signed-off-by` so I couldn't take it directly.

PR #2354, opened 2018-09-19
Set HBB and HBI partitions' access to Read Only

popfuture:
This change removes a revert of the "Mark Read Only Partitions as
Such" change and updates the pnor makefiles to ensure HBB and HBI
partitions are marked read only at the pnor level.

Signed-off-by: Jaymes Wilks <mjwilks@us.ibm.com>


stewart-ibm:
retest this please

PR #2388, opened 2018-10-08
In P8, use pnor layout with all possible sections as read only

popfuture:
Signed-off-by: Jaymes Wilks <mjwilks@us.ibm.com>


wghoffa:
Comparable changes already merged here: https://github.com/open-power/op-build/pull/2387

PR #2395, opened 2018-10-11
petitboot: Bump version to v1.7.3

sammj:
Includes a minor fix to the network requery feature


stewart-ibm:
I've cherry picked the travis and CI fixes in with https://github.com/open-power/op-build/pull/2396 so we should all be good

PR #2404, opened 2018-10-16
Update openpower-pnor for signed VERSION partition

sammj:
Bring openpower-pnor up to date for the op-build-2.0.y branch. This allows the VERSION partition to be signed and also brings along some quality of life changes that have happened since op-build-2.0.


sammj:
```
Commit sha: 6668eb1, Author: Nick Bofferding, Committer: Samuel Mendoza-Jonas; Expected "Nick Bofferding opensource@bofferding.net", but got "Nick Bofferding bofferdn@us.ibm.com".
```
Ah the DCO check failed for patch 4 since Nick used a different author/signed-off-by combo - that would be mismatched in the original commit too though.

PR #2435, opened 2018-11-02
Use latest Hostboot rather than old stable

stewart-ibm:
We need this for HIOMAP as well as signed VERSION partition support,
amongst other things.

While not necessarily ideal on all fronts, it's what we have now.


stewart-ibm:
@wghoffa what do you think on this? It removes the witherspoon_dev defconfig, which beraks the CI build in Austin (but we can fix that by removing building the defconfig).

dcrowell77:
> We need this for...
Who is the 'we' here?  Why can't the _dev be used?

Would it be better to move BR2_HOSTBOOT_STABLE_VERSION up instead?  I suppose we would have to convince ourselves that recent levels are in fact 'stable' but it would keep op-build from being a moving target (which seemed to have been the goal of the STABLE/LATEST to being with).

I'm not actually opposed to the switch fundamentally, just wondering if this is the right way to go.

wghoffa:
I missed the meeting on Halloween -- so not sure what all was decided as far as how/when to release code. I kind of like the idea of updating Stable with the latest HB commits though then I wonder what the use-case for Latest would be.

stewart-ibm:
I think the argument of switching to LATEST rather than updating the STABLE is that there is no hostboot release or stable version, it's just a snapshot of the tree at some point in time that there's a general feeling that it's kind-of okay.

wghoffa:
@stewart-ibm -- For now I went with a one time move up to get things flowing again:

https://github.com/open-power/op-build/pull/2450. We can certainly discuss getting rid of the LATEST/STABLE content altogether (which agreed would allow us to drop witherspoon_dev)

stewart-ibm:
Okay, I'll close this PR then as we have the net effect of it in that PR

PR #2480, opened 2018-11-22
Add p9_simics_defconfig

rarbab:
Add a minimal config target allowing us to test the default p9 configuration in SIMICS.


stewart-ibm:
ok to test

rarbab:
Oops. I left a bug in here. Currently, this error is always going to trip:

devicetree.mk:17: *** No device tree source set. Check your BR2_DEVICETREE_SOURCE setting.  Stop.

I need to wrap that so the check only happens when BR2_PACKAGE_DEVICETREE=y. Will fix!

PR #2483, opened 2018-11-22
Add p9_simics_defconfig

rarbab:
Add a minimal config target allowing us to test the default p9 configuration in SIMICS.


shenki:
The packaging looks good.

Given the only addition is the dtb in the output directory, how about we add this to the mambo defconfig (we could rename it to something else if you have a good suggestion)?

rarbab:
Well, the new config does differ more than just adding a dtb. You're right, it may well be appropriate to add a dtb to the mambo defconfig, but I do also like the idea of a new config that's very minimal and plain (BR2_LINUX_KERNEL_DEFCONFIG="ppc64le", etc.) like the one here, for basic chip development.

I don't feel too strongly either way. @apopple, you are a likely user of this type of defconfig--any input?

As far as naming, it's kind of dictated by the device tree contents, which specifically list p9/simics.

shenki:
> Well, the new config does differ more than just adding a dtb (kernel config, binary formats, more). You're right, it may well be appropriate to add a dtb to the mambo defconfig, but I do also like the idea of a new config that's very minimal and plain (BR2_LINUX_KERNEL_DEFCONFIG="ppc64le", etc.) like the one here, for basic chip development.

> 
> I don't feel too strongly either way. @apopple, you are a likely user of this type of defconfig--any input?
> 
> As far as naming, it's kind of dictated by the device tree contents, which specifically list p9/simics. Maybe openpower_simics_defconfig?

rarbab:
> FWIW, the skiroot defconfig has even less enabled than the ppc64 one:

Ah, I didn't know that! Thanks.

Okay, I think I'm on board with your original suggestion now, just adding a devicetree to openpower_mambo_defconfig. I guess we should get all the naming straight, though. The .dts file I'm adding has a `simics {};` node, which makes me want to call the file `p9_simics.dts`, which makes me want to call the whole config `openpower_simics_defconfig`. What do you think?

shenki:
> > FWIW, the skiroot defconfig has even less enabled than the ppc64 one:
> 
> Ah, I didn't know that! Thanks.
> 
> Okay, I think I'm on board with your original suggestion now, just adding a devicetree to openpower_mambo_defconfig. I guess we should get all the naming straight, though. The .dts file I'm adding has a `simics {};` node, which makes me want to call the file `p9_simics.dts`, which makes me want to call the whole config `openpower_simics_defconfig`. What do you think?

Is it simics specific? The device tree looks like it could be used for things such as mambo too.

If that's the case we could call it p9_sim.dts or similar.

We could rename the defconfig to be opal_defconfig (or similar), where we build:
 * Linux kenrel
 * skiboot
 * dtb for simulators

I say dtb as we might end up adding device trees for other chips too.

rarbab:
> Is it simics specific? The device tree looks like it could be used for things such as mambo too.

Yes. Apparantly skiboot uses these dt nodes to determine which sim we're in:

        /* Detect mambo chip */
        if (dt_find_by_path(dt_root, "/mambo")) {

        /* Detect simics */
        if (dt_find_by_path(dt_root, "/simics")) {

        /* Detect Awan emulator */
        if (dt_find_by_path(dt_root, "/awan")) {

        /* Detect Qemu */
        if (dt_node_is_compatible(dt_root, "qemu,powernv")) {

> We could rename the defconfig to be opal_defconfig (or similar), where we build:
> 
>     * Linux kenrel
>     * skiboot
>     * dtb for simulators
> 
> I say dtb as we might end up adding device trees for other chips too.

That will require some enhancement as the current devicetree.mk assumes there's only one output.

stewart-ibm:
ok to test

shenki:
Following discussion, it was decided we would like to have device trees for both mambo and simics in the skiboot tree.

The kernel was considered but given some uses care about skiboot more than the kernel, and Linux will always be booted with skiboot, skiboot makes more sense.

rarbab:
> Following discussion, it was decided we would like to have device trees for both mambo and simics in the skiboot tree.

That does seem like a good place for them to live. I'll go ahead and close this pull request.

Note that when skiboot does produce device trees, we will still probably have to add some kind of op-build config option for defconfigs that want them included in output/images.

PR #2494, opened 2018-11-28
Change to update hdat pnor partition in openpower package

sampmisr:
This is to include the hdat data in the open power pnor partition.


sampmisr:
Identified some more changes in this area. The work is going on. We will submit all the changes together.

PR #2498, opened 2018-11-30
RFC: Towards ffspart rather than much perl

stewart-ibm:
The way we currently generate the PNOR image itself is rather convoluted, slow, and spread across at least three git repositories.

Doing something simple like "this partition should be signed" is overly complex, and build times for a PNOR are much longer than they should be for something which is effectively copying 64MB of data from a bunch of files into one file.

So, we have had the `ffspart` utility sitting in skiboot for a while. It's about 400 lines of C that uses the same libflash infrastructure that skiboot and pflash use. It takes a CSV input and constructs a FFS layed out PNOR image.

In order to ensure we don't regress when transitioning over, this PR builds two pnor images: one using the existing method and the other taking the XML, converting it to the CSV and then producing a PNOR which we diff against the traditional method.

The next step would be to produce the ffspart based image independently of the XML method and diff it, before finally removing the old method.

This is more of a RFC to see how it all fairs in the test matrix that CI builds.


stewartsmith:
Any Cs for this RFC?

dcrowell77:
I'm definitely not going to argue that the current process is great.  If the laid out steps work then I'd entertain the change.

PR #2513, opened 2018-12-11
Random docs improvements

stewart-ibm:
A bunch of misc doc improvements that should be pretty uncontroversial.

@shenki I copied over the wiki page for KernelProcess and updated it a bit. What are your thoughts on maintaining that here rather than wiki? It does mean we can update it as we bump kernel version.

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -7,30 +7,40 @@ machines.
 Firmware Components
 -------------------
 
-Buildroot
+Buildroot_
   We use http://buildroot.net/ as the build system for assembling a firmware
   image. `op-build` is a "Buildroot overlay". We build a kernel and initramfs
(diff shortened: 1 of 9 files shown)


shenki:
I like the wiki as it lowers the barrier to adding/updating/correcting documentation and contributors do not need to wait for anyone in order to fix it.

If you would prefer it to be in the tree then I guess we can live with that.

stewart-ibm:
> I like the wiki as it lowers the barrier to adding/updating/correcting documentation and contributors do not need to wait for anyone in order to fix it.
>
> If you would prefer it to be in the tree then I guess we can live with
> that.

I only prefer it because we have other process/maintenance things
documented (and should be increasing that).

-- 
Stewart Smith
OPAL Architect, IBM.

PR #2516, opened 2018-12-11
vesnin: Add driver and firmware for QLogic devices

artemsen:
Support for QLogic Corp. FastLinQ QL41000 Series 10/25/40/50GbE Controller.
Skiboot's Linux Kernel version is usually newer than linux-firmware
package from buildroot, but modifying the buildroot's script is not
a trivial task because it exists as submodule for op-build project.
So, firmware of the network controller for current Linux kernel version
is placed directly to the filesystem overlay in our repository.


stewart-ibm:
ok to test

shenki:
> Support for QLogic Corp. FastLinQ QL41000 Series 10/25/40/50GbE Controller.
> So, firmware of the network controller for current Linux kernel version
> is placed directly to the filesystem overlay in our repository.

It looks like you forgot to add the firmware file to this commit.

Looking at the qed driver in 4.19, the firmware version is 8.37.2. This version is present in the version of linux-firmware that is in buildroot 2018.11. I suggest updating the firmware-whitelist to specify the precise file you want, otherwise you will end up with all of the .bin files in your filesystem.

Normally when a request comes in to support another adaptor we add it to the common kernel configuration. Is there any reason you don't want to do that here?

openpower/configs/linux/vesnin_fragment
```diff
@@ -0,0 +1,2 @@
+CONFIG_QED=m
+CONFIG_QEDE=m
```
shenki: I suggest putting this in `openpower/configs/linux/skiroot_defconfig`

openpower/scripts/firmware-whitelist
```diff
@@ -10,6 +10,7 @@ whitelist=(     'acenic/'
                 'cxgb4/t4fw-1.16.63.0.bin'
                 'cxgb4/t4fw.bin'
                 'cxgb3/'
+                'qed/'
```
shenki: Instead of including all of the qed directory, specify the file that is required by the current kernel (4.19). This looks to be 8.37.2.0, so we want `qed/qed_init_values_zipped-8.37.2.0.bin`.

artemsen: Yes, it is. But the problem is in buildroot, it still use 8.37.2.0:

buildroot/package/linux-firmware/linux-firmware.mk:
```
ifeq ($(BR2_PACKAGE_LINUX_FIRMWARE_QLOGIC_4X),y)
LINUX_FIRMWARE_FILES += \
	qed/qed_init_values_zipped-8.33.11.0.bin
endif
```

artemsen: Of course, I can set LINUX_FIRMWARE_FILES through environment variables, but it doesn't resolve the problem when buildroot contains old linux-firmware package without firmware defined for current kernel. Force set of LINUX_FIRMWARE_VERSION also cannot be used because I have to modify the hash file of linux-firmware that belongs to buildroot.

openpower/configs/linux/vesnin_fragment
artemsen: Ok, I will put them into common config.
But we have to decide what we will do with firmware files, I found out that the Linux kernel we use sometimes is newer than linux-firmware package in buildroot (see below).

openpower/scripts/firmware-whitelist
shenki: https://github.com/open-power/op-build/pull/2524

AlexanderAmelkin:
@stewart-ibm, can we expect a merge soon, please?

stewart-ibm:
Alexander Amelkin <notifications@github.com> writes:
> @stewart-ibm, can we expect a merge soon, please?

Sure! I'm back from the holidays and at least starting to merge things again.

-- 
Stewart Smith
OPAL Architect, IBM.

shenki:
retest this please

stewart-ibm:
retest this please

PR #2520, opened 2018-12-12
Add op-test-framework v0.13 as a git-submodule

ghost:
The idea is we tie an op-build commit to a op-test commit so that we
control what tests people run in CI via code in op-build repo rather
than random jenkins jobs.

It's also a point that an op-build commit should (must!) pass the tests
in the given op-test commit.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = buildroot
 	branch = 2018.08-op-build
 	url = https://github.com/open-power/buildroot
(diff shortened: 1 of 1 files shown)


stewart-ibm:
So, I'm wanting this here (and to get it merged) as a step towards having all the tests that should pass be alongside tho code.

This also means that if we want to update what op-test is run, it comes through the normal CI.

PR #2524, opened 2018-12-13
buildroot: Pull in fix for LINUX_FIRMWARE_QLOGIC_4X

shenki:
This updates linux-firmware package to pull in all firmware versions for
QLOGIC 4X ("qed"). This firmware is used by Vensin.


shenki:
We've since bumped buildroot for other fixes, so this is part of the tree now.

PR #2531, opened 2018-12-19
vesnin: Add temperature sensor support for CPU cores

artemsen:
IPMI protocol imposes limits on numbers of sensors identifiers (1 byte).
This limitation doesn't allow us to use unique temperature sensor ID for
each CPU core (4 CPU * 12 cores = 48 sensors).
Added patch makes OCC use own ID as channel label instead of using IPMI
temperature sensor ID.

### DIFF
--- a/openpower/patches/vesnin-patches/occ-p8/occ-0002-Add-temperature-sensor-support-for-CPU-cores.patch
+++ b/openpower/patches/vesnin-patches/occ-p8/occ-0002-Add-temperature-sensor-support-for-CPU-cores.patch
@@ -1,0 +1,43 @@
+From 0ec6ab11dcec304876802d1ed7bb01bd059ae311 Mon Sep 17 00:00:00 2001
+From: Artem Senichev <a.senichev@yadro.com>
+Date: Tue, 18 Dec 2018 14:16:56 +0300
+Subject: [PATCH] Add temperature sensor support for CPU cores
+
+IPMI protocol imposes limits on numbers of sensors identifiers (1 byte).
+This limitation doesn't allow us to use unique temperature sensor ID for
+each CPU core (4 CPU * 12 cores = 48 sensors).
+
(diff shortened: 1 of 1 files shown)


stewart-ibm:
ok to test

marthabroyles:
The sensor IDs come from the xml so they match with the BMC, you can't arbitrary assign sensor ids the BMC will not recognize the sensor ID you are proposing.  Sensor IDs need to be changed in the xml.

artemsen:
I wish to add them to the XML, but IPMI supports only 255 sensors (1 byte field), for 4-sockets machine it's impossible to use real unique ID - we have 128 DIMMs on 16 memory buffers and 4 12-core CPUs. On BMC side we have static ini-file for hwmon and this file is not autogenerated from MRW.

marthabroyles:
As long as this is only pulled in for builds that are specifically for 4 socket with BMC that will recognize the same mapping that is fine.  We will not be merging this change back into our master OCC branch which supports multiple systems and BMCs.

artemsen:
Yes, of course. This patch is not intended to be merged into OCC master, it is a standalone patch for VESNIN server only, so I placed it into our patches dir inside op-build project.

stewart-ibm:
Martha Broyles <notifications@github.com> writes:
> As long as this is only pulled in for builds that are specifically for
> 4 socket with BMC that will recognize the same mapping that is fine.
> We will not be merging this change back into our master OCC branch
> which supports multiple systems and BMCs.

You wouldn't want it as a build time option in the OCC code?

If a variety of platforms need to carry OCC patches in op-build, this
could prevent you from updating OCC code in the future if those patches
ever fail to apply to updated OCC code. It possibly also makes it harder
to maintain OCC code into the future, as there'd no longer be one
central repository.

-- 
Stewart Smith
OPAL Architect, IBM.

PR #2557, opened 2019-01-14
Fix RC09 ODT errors on Nimbus Dual Drop Configurations

wghoffa:
- Remove no longer needed LRDIMM Revert Patch
  - Add fix for RC09 ODT errors as patch


dcrowell77:
P9 CI fail is the known skiroot issue with the PCI slot labels on Boston
P8 CI fail is the known issue with downlevel palmetto

I think we can go ahead and merge this.

PR #2597, opened 2019-02-03
Kernel 4.20.6

shenki:
Update to Linux 4.20.6, via 4.20.3. Includes buildroot bump to support 4.20 and a CI fix for Fedora.


shenki:
Replaces https://github.com/open-power/op-build/pull/2571

PR #2631, opened 2019-02-18
petitboot: Reset sudo timeout & update prompt

sammj:
Call 'sudo -K' on shell exit to remove cached credentials. This avoids
the case where a user could still automatically sudo after setting a
password if they had entered the shell recently.
Also explicitly set the PS1 based on the current UID so that 'su' by
default still updates the prompt.


sammj:
retest this please

PR #2642, opened 2019-02-20
Add BR2_SKIBOOT_DEVICETREE

rarbab:
Skiboot can build device trees now. Let's deliver them in a freshened simulator-friendly defconfig.


stewart-ibm:
ok to test

rarbab:
Note that this pull causes a failed CI check of CONFIG=openpower_mambo, because it actually changes the name of that config.

That will need to be addressed in the CI if/when this gets merged.

shenki:
retest this please

rarbab:
> We could just make building the dtb the default. If you agree, I suggest we do that in a future patch.

I don't really have a preference, but it seems like users of this new opal_defconfig would be the only ones who actually need device trees.

Also, looks like I've got a merge conflict now because BR2_LINUX_KERNEL_CUSTOM_VERSION_VALUE has been updated upstream. Should I fix that or can it be handled at merge?

stewart-ibm:
retest this please

stewart-ibm:
retest this please

PR #2643, opened 2019-02-20
petitboot: Update login prompt message

sammj:
If no password is set Petitboot will automatically open a shell as the
root user, and prints a message. Update this message to be more helpful
to a user if they're not aware that password authentication is
available.


sammj:
When not expecting it some users found the "elevating shell" message a bit alarming so let's change it to something more useful - bikeshed away on the message itself if we can make it more intuitive.

stewart-ibm:
I think this bike shed is the appropriate colour scheme for an Australian Summer.

PR #2711, opened 2019-03-19
add p8dtu config files an patches.

jimsmc:


openpower/patches/p8dtu-patches/linux/0001-VGA-handshake-with-BMC.patch
```diff
@@ -0,0 +1,68 @@
+From 4c5f6ff7a2b7c2df17d16c3ce18b19036be95223 Mon Sep 17 00:00:00 2001
+From: Jim Yuan <jim.yuan@supermicro.com>
+Date: Mon, 17 Dec 2018 15:57:05 -0800
+Subject: [PATCH] VGA handshake with BMC
```
stewart-ibm: @shenki @amboar Are we any closer for a standard way to do this across openpower machines?

@shenki I'm gathering we don't particularly want to carry a platform specific kernel patch forever?

shenki: We don't want to carry a platform specific patch ever.

jk-ozlabs: Yep, we have a standard way to do that, which doesn't require any host-side code. It's simply implemented by having the BMC watch for the host side setting the VGA password register after power-on. This means that the VGA driver is actively controlling the VGA device, and is all the signaling we should need.

@jimsmc if this design isn't sufficient, we'd need to know why before justifying a custom implementation of the VGA handshake.

shenki:
@jimsmc what was the resolution here? It's always a good idea to comment why you're closing a bug when there's outstanding discussion.

jimsmc:
I forwarded this question to BMC firmware team and no resolution yet.

PR #2774, opened 2019-04-10
Retarget op-build to sb-signing-utils package v0.8

mtvaught-ibm:
Point to sb-signing-utils package v0.7 which fixes a bug when running OpenSSL v1.1.0 or higher


stewart-ibm:
Is this something that should also go to the op-build v2.3 release?

mtvaught-ibm:
If possible, yes. I know the issue it fixes is causing some build failures.

fspbld:
Yes, anywhere the new buildroot exists (ie, 2019xxxx) it will be needed because the newer openssl that is referenced in that buildroot level is what exposes the error in the signing-utils package.

stewart-ibm:
Matthew Vaught <notifications@github.com> writes:
> If possible, yes. I know the issue it fixes is causing some build
> failures.

https://github.com/open-power/op-build/pull/2780 will bring it in for
2.3.y (and thus the upcoming IBM OP930 release).

I'm trying to remember if we've bumped the buildroot in 2.0.y, I don't
*think* so at least...

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
Pam Eggler <notifications@github.com> writes:
> Yes, anywhere the new buildroot exists (ie, 2019xxxx) it will be
> needed because the newer openssl that is referenced in that buildroot
> level is what exposes the error in the signing-utils package.

Okay... We could just bump it in 2.0.y to preempt the problem.

-- 
Stewart Smith
OPAL Architect, IBM.

PR #2778, opened 2019-04-11
linux: Disable XMON by default

shenki:
This changes the kernel option so that xmon is not initialised by
default. It can be re-enabled for debugging by adding xmon=on to the
petitboot kernel's command line:

 nvram -p ibm,skiboot --update-config bootargs="xmon=on"


shenki:
@mpe

daxtens:
Putting my future secure boot goggles on: is the magic sysrq key or /proc/sysrq-trigger in petitboot? I'm just thinking that even with this patch, a root user on a host could then write to an insecure nvram variable and then be able to use xmon to break kernel integrity.

Perhaps we could mitigate this with the upcoming xmon read only support (http://patchwork.ozlabs.org/patch/1080551/), but beyond the ease of debugging, is there any reason not to just compile out xmon entirely?

mpe:
There is a class of bugs where having xmon is very helpful, as long as the person debugging knows how to drive it. So having it available is a good thing IMO.

But I agree that by default it's more helpful for most people for the petitkernel to just oops and reboot. So I think this change is good.

I think the interaction with secure boot should be a separate discussion, not all people will be using secure boot capable systems, or booting them in secure mode, so removing xmon entirely would be overkill I think.

daxtens:
OK, I think that's a totally reasonable position. We'll need some patches for SB support anyway - in those we'll either enforce XMON_RO if it lands or just disable it completely.

Acked-by: Daniel Axtens <dja@axtens.net>

sammj:
Looks good. This reminds me that we really need a better way of setting NVRAM options from the BMC for example without hand-crafting the partition ourselves - making re-enabling this if we need easier.

daxtens:
@sammj so perhaps you would like nvram to be accessible via some sort of
generic firmware variable file system?

sammj:
@daxtens well if your generic firmware variable file system runs on the BMC and knows how to deal with VPNOR.. yes! :D

daxtens:
What tools do we currently use to talk VPNOR? I know of nvram in userspace
for pseries guests; is it the same?

sammj:
@daxtens Only `nvram` at the moment, which works on powernv as well. However what I'm talking about is writing to NVRAM without the host - eg. if it's crashing before we get to the Petitboot shell. In that case we need to modify it from the BMC and no BMC implementations currently has an `nvram` equivalent.

PR #2803, opened 2019-04-17
Configuration updates to compile with GCC 8

LF-Luis:


stewart-ibm:
ok to test

madscientist159:
Unfortunately this patch also yields a system that will not IPL, with no hostboot console output (https://github.com/openbmc/openbmc/issues/3531).  Reverting buildroot to GCC6 allows hostboot to start but it then immediately fails in targeting (https://github.com/open-power/hostboot/issues/173)

bofferdn:
> Fixes for OCC gcc8 compile coming separately via occ package
> Fix for crash in HB coming separately via hostboot package
> Fix for PNOR layout changes to fit BOOTKERNEL for p8/p9 coming via pnor package
This will have to be resubmitted to bring all those in.  Good news is we built whole driver in GCC8 and booted on a WSP, so should be out of the woods once we link in all these changes.

bofferdn:
I attemped to build this patch on master on a RHEL 7 machine w/ witherspoon defconfig and it passed; I don't get why it's giving inconsistent results vs. local testing.

lpfluisfer:
retest this please

bofferdn:
retest this please

lpfluisfer:
retest this please

sammj:
I'm assuming this PR is for testing only and the included patches will go to their relevant repositories instead?

The Linux Makefile patch is interesting, @shenki can we handle that with a config option instead?

This should probably update the other defconfigs as well, not just P9 platforms.

shenki:
The kernel changes are a non-starter. If we end up requiring this we will upstream them to the powerpc or top level makefile.

lpfluisfer:
Running CI without the Kernel changes to see if new sizing still holds.

sammj:
```
ERROR: PnorUtils::checkSpaceConstraints: Image provided (/tmp/hostboot-jenkins.swg-devops/jenkins-slave/workspace/Hostboot/OpenPower/GHPR/ubuntu-ppc64le-2242/CONFIG/witherspoon/OS/Ubuntu_18.04/build/op-build/output/host/powerpc64le-buildroot-linux-gnu/sysroot/openpower_pnor_scratch//zImage.epapr) has size (32505856) which is greater than allocated space (16252928) for section=BOOTKERNEL.  Aborting! at /tmp/hostboot-jenkins.swg-devops/jenkins-slave/workspace/Hostboot/OpenPower/GHPR/ubuntu-ppc64le-2242/CONFIG/witherspoon
```
Appears to have worked for some platforms but not others - that's 2x the normal size though I suspect something else is up there.

madscientist159:
@sammj 2x the normal size is the scripting breaking down -- if you're even 1 byte over, dd spits out another block since the block size is set to the allowed partition size.  Presto, double size for a slight overage.

It made debugging fun for a while until I figured out what was going on. :smile:

sammj:
@madscientist159 Haha, that does ring a bell. Ugh, my kingdom to find time to finish off the `ffspart` rework.

bofferdn:
@shenki Can we initiate the process to pull the kernel changes to upstream.  This is the only  way after a month of trying we have been able to fit everything in the space constraint.  Otherwise, we'll need help to somehow prune the content down or something.

shenki:
> @shenki Can we initiate the process to pull the kernel changes to upstream. This is the only way after a month of trying we have been able to fit everything in the space constraint. Otherwise, we'll need help to somehow prune the content down or something.

Yes. Please hit me up on slack if you need instructions on how to send kernel patches.

stewart-ibm:
https://github.com/open-power/op-build/pull/3004 removes the P8 platforms, so that solves the P8 problem.

klauskiwi:
The ci/build* scrips have changed significantly since this was first proposed, and there are plenty of other PRs (here and on Hostboot, OCC etc) proposing solutions for compilation with modern compilers.. so far, looks like the theory of 'removing power8 entirely' is winning (#3145).

I'll politely close this, but if there's any aspect of this that we should still discuss, feel free to reopen/speak-up

LF-Luis:
Thank you for closing this.

PR #2794, opened 2019-04-15
2.3.y kernel 5.0.7

stewart-ibm:
Only merge after https://github.com/open-power/op-build/pull/2793 and https://github.com/open-power/op-build/pull/2792

This PR is all early so that we get a headstart on CI

PR #2808, opened 2019-04-18
2.3.y kernel 5.0.7

stewart-ibm:


stewart-ibm:
Like https://github.com/open-power/op-build/pull/2794 but for the correct branch this time.

(the CI has already passed for this in https://github.com/open-power/op-build/pull/2794 - although the status of that is overwritten now)

PR #2851, opened 2019-05-06
Do Not Merge: Testing changes with CI

lpfluisfer:
There is a patch for PNOR I have pending review. In the meanwhile, I am testing those changes by introducing a patch. This PR is not meant to be merged.


bofferdn:
oh this special PR would have had to also specify the most recent Hostboot and OCC commits as well, that is why it failed

PR #2845, opened 2019-05-06
kernel: Move to Linux v5.0.13-openpower1

shenki:
Signed-off-by: Joel Stanley <joel@jms.id.au>

PR #2857, opened 2019-05-08
linux: Re-enable radix MMU

shenki:
A workaround was put in place for the issue exposed by STRICT_KERNEL_RWX
which meant RADIX required HUGETLBFS. As we didn't have HUGETLBFS
enabled the petitkernel defaulted to radix.

This enables HUGETLBFS so we go back to using the radix MMU.

The disabling of STRICT_KERNEL_RWX can be reverted with this alternate
workaround in place.

See https://github.com/linuxppc/issues/issues/237.

Fixes: 885ef25a0d68 ("linux: temporarily disable STRICT_KERNEL_RWX")


shenki:
I made a mistake, we are not yet shipping the kconfig change that would see us defaulting to hash. I have reworked the commit message and added this change to the v5.0.13 PR which does include the change: https://github.com/open-power/op-build/pull/2845

PR #2868, opened 2019-05-10
skiboot: Fix building device tree with host-dtc

shenki:
We need to do the build with the PATH set to the host tools. Without
this change the build fails when the host lacks a dtc, as it is not in
the PATH.


shenki:
Good question. I think it was due to a combination of luck, and only stewart's CI trying to build with SKIBOOT_DEVICETREE enabled. I updated the commit message with some more details.

PR #2878, opened 2019-05-14
linux: Work around broken OCC/powersave

shenki:
There is a bug in current Power9 firmware where Linux appars to crash
after running for some amount of time. There are no obvous symptoms in
the skiboot msglog, or dmesg, but CPUs are shown to be stuck in the
HOMER region, part of the OCC's power management code.

Vaidy said:
> This is power management issue. Hostboot using different stop-api vs us
> in OPAL may cause issues.  As such the srtop-api and HCODE image are
> backward compatible.  But what we are discovering is that hostboot
> modified HOMER before us and there is some fallout

This can be worked around by disabling stop4 and stop5 in skiboot, or in
the kernel by setting powersave=off.

This is temporary until the OCC issue is resolved.

### DIFF
--- a/openpower/configs/linux/skiroot_defconfig
+++ b/openpower/configs/linux/skiroot_defconfig
@@ -53,7 +53,7 @@ CONFIG_NUMA=y
 CONFIG_PPC_64K_PAGES=y
 CONFIG_SCHED_SMT=y
 CONFIG_CMDLINE_BOOL=y
-CONFIG_CMDLINE="console=tty0 console=hvc0 ipr.fast_reboot=1 quiet"
+CONFIG_CMDLINE="console=tty0 console=hvc0 ipr.fast_reboot=1 quiet powersave=off"
 # CONFIG_SECCOMP is not set
 # CONFIG_PPC_MEM_KEYS is not set
 CONFIG_NET=y


stewart-ibm:
How about instead we revert back the offending HCODE to something that remotely works?

dcrowell77:
I thought we solved the stop-api issue already with updated opal code.  Is this something new?

svaidy:
@dcrowell77 The issue seem to be build/integration and not resolved by updating stop-api in OPAL.  While OPAL and hostboot need to use same level of stop-api. That is not the breakage here.

All parts of the code and binary is not pushed out to various trees under op-build so that a correct hcode image is built.

By design hcode image is backward compatible, you can use older stop-api on newer hcode image. We had removed the need for this co-dependency last year.  However the code and binary flow from developer to EKB and to public repositories are still broken and needs attention and fixing.

svaidy:
@shenki System is able to get past petitiboot and get to host. Hence as a workaround we can provide powersave=off to kernel cmdline in petitboot and in the installed distros for a few more days until this hcode issue can be fixed.  

We will need this code change/PR merged only if we break so early in the boot and unable to get to petitboot or host kernel at all.

stewart-ibm:
Let's revert the HCODE and any offending stop-api changes in hostboot.

We cannot have this kind of mess and co-dependencies that break in such a fundamental way rather than degrading gracefully.

dcrowell77:
> Let's revert ... any offending stop-api changes in hostboot.

That is easier said than done.  The PM team would need to identify the problems and make changes on their side to push through.  I've heard it might just be a code delivery issue.

stewart-ibm:
Yeah, they need to go do that then. They cause the pain, they own the pain :)

Re-adding `powersave=off` is a solid NAK from me.

svaidy:
Unfortunately, rolling forward HCODE image is much easier than rolling back.  Current op-build master has hw051619a.940 
https://github.com/open-power/op-build/commit/a658ffe9e4245116feaa67cfd43ce3df549f1bed

This hcode image has all necessary fixes to the fixes :).  Prem has reviewed the hcode build that came off my op-build master branch today.  At this time we have only verified the build. We will find a way to test this image and report.

svaidy:
Tested new hcode image on witherspoon with upstream op-build:
https://github.com/open-power/op-build/commit/b6bbfeb822ccce10a5e8d414aec670eef9e49480

open-power-witherspoon-v2.3-rc2-124-gb6bbfeb8
        buildroot-2019.02.2-20-g9a0714da65
        skiboot-v6.3-57-g76f7316bc8fc
        hostboot-06a885d
        occ-4b93968
        linux-5.1.2-openpower1-p9c63c66
        petitboot-v1.10.3
        machine-xml-a6f4df3
        hostboot-binaries-hw043019a.940
        capp-ucode-p9-dd2-v4
        sbe-c54904f
        hcode-hw051919a.940

hcode-hw051919a.940 has all necessary fixes and all stop states are working find with this build.

PR #2898, opened 2019-05-27
busybox: Enable HTTPS/FTPS support for wget

artemsen:
Allows using internal TLS implementation to get kernel/initrd images
over HTTPS or FTPS protocols.


stewart-ibm:
ok to test

stewart-ibm:
I'm guessing the big size increase would come if we were including CA certificates rather than just the busybox wget https code. I know we've been getting a bit closer on image size recently, but let's see how it goes!

artemsen:
Not only CA, we also need the 'real' wget. Busybox doesn't support certificate validation at all.
And we have to close this vulnerability sooner or later.

P.S. This patch increases the size of busybox by only 16 bytes.

stewart-ibm:
Artem Senichev <notifications@github.com> writes:
> Not only CA, we also need the 'real' wget. Busybox doesn't support certificate validation at all.
> And we have to close this vulnerability sooner or later.

Ahh. I guess it would be a good idea to.

More of a gotcha than a vuln though.

It'd probably be good to have decent and extensive end user docs on all
the features and associated risks.

> P.S. This patch increases the size of busybox by only 16 bytes.

That's just magic. Wow.

-- 
Stewart Smith
OPAL Architect, IBM.

PR #2906, opened 2019-05-30
Point op-build P8 hostboot at commit to report cache-count-disabled O…

bofferdn:
Points OP build at the P8 hostboot package commit which enables reporting to the OS
that the cache-count-disabled Spectre workaround is available.

Signed-off-by: Nick Bofferding <opensource@bofferding.net>


dcrowell77:
Looks right to me

stewart-ibm:
Do we have any p8 platforms building out of 2.0.y ? I don't *think* so, but it might be worth having this head there too in case someone else is.

PR #2907, opened 2019-05-31
[2.0.y] Bump p9dsu-xml to 734a35ebe5af49dba57269ca26065d40b876c962

youhour:
Bump p9dsu-xml to 734a35ebe5af49dba57269ca26065d40b876c962 to pickup the following fixes:

734a35e - careyz@supermicro.com - 2019-04-05 - MEMD:VM keyword is not set in system vpd on P9DSU-C
e1a56eb - careyz@supermicro.com - 2019-03-28 - Checkstop in istep 14.1 (mss_memdiag) - dimm garded
218a77a - jimsmc - 2018-05-03 - Fix duplicate capps target

PR #2909, opened 2019-06-02
[2.3.y] Bump p9dsu-xml to 734a35ebe5af49dba57269ca26065d40b876c962

youhour:
Bump p9dsu-xml to 734a35ebe5af49dba57269ca26065d40b876c962 to pickup the following fixes:
   MEMD:VM keyword is not set in system vpd on P9DSU-C
   Checkstop in istep 14.1 (mss_memdiag) - dimm garded
   Fix duplicate capps target


mikey:
Does this supersede PR #2907 ? If so how and where are the descriptions @sammj asked for?

youhour:
> 
> 
> Same feedback as for #2907
> 
> Is this backporting a commit from master? If so, please ensure that commit is mentioned (i.e. git cherry-pick -xs or use stable tools to have "[abcdef1235 upstream]" at top of commit message).
> 
No we cannot backport from master since the same file for master contains different contents.

> I really don't want us to have references to private bug trackers is our commits. Doing so tends to make people lazy in describing the problem and the solution in the commit message.
> 
> > Pickup the latest p9sdu-xml changes
> 
> "latest" is subjective and will be incorrect in the future. Better to say "Bump p9dsu-xml to 734a35ebe5af"
OK.  

> 
> > Fix issue 1414: MEMD:VM keyword is not set in system vpd on P9DSU-C
> 
> Don't mention issue number, one would assume it's an op-build issue number, which it isn't.
> 
> This isn't the best description of the problem/solution either, but I'm guessing the p9dsu-xml commit isn't any better either.
OK.

> 
> > Fix issue 1381: Checkstop in istep 14.1 (mss_memdiag) - dimm garded
> 
> Again, don't mention issue number.
> What changes to avoid this?
OK.

> 
> > Fix duplicate capps target
> 
> okay. I'm not sure of the impact of this, but that's not a big concern.
> 
> > bug: issue1414 issue1381
> 
> The correct format for pointing to bugs that are fixed is using the "Fixes: URL/bug-number" notation, one per line.
May be it is best not to include this...

> 
> > branch: op-build-2.0.y
> 
> This is not needed. If the commit is being cherry-picked back from master (which I'm guessing it should be, unless these bugs are specific to 2.0.y and 2.3.y, then have the "cherry-picked from abcd1234" format as provided by git cherry-pick -xs.
It is not a cherry-pick from master because the content of openpower/configs/p9dsu_defconfig is different between op-build-2.0.y, op-build-2.3.y and master.

stewart-ibm:
Michael Lim <notifications@github.com> writes:

>> > branch: op-build-2.0.y
>> 
>> This is not needed. If the commit is being cherry-picked back from master (which I'm guessing it should be, unless these bugs are specific to 2.0.y and 2.3.y, then have the "cherry-picked from abcd1234" format as provided by git cherry-pick -xs.
> It is not a cherry-pick from master because the content of
> openpower/configs/p9dsu_defconfig is different between op-build-2.0.y,
> op-build-2.3.y and master.

So p9dsu-xml has forked for 2.3.y and for 2.0.y and then there's master
as well? Why?

-- 
Stewart Smith
OPAL Architect, IBM.

stewart-ibm:
As per https://github.com/open-power/op-build/pull/2907, gathering this is no longer needed and closing

PR #2931, opened 2019-06-11
buildroot: Move to 2019.02.3 stable release

shenki:
Upstream release notes:

 http://git.buildroot.net/buildroot/plain/CHANGES?id=2019.02.3
 http://lists.busybox.net/pipermail/buildroot/2019-June/252046.html


shenki:
We should be going straight to 2019.05, but as that is taking some time, this allows us to pick up the latest stable release of 2019.02.

PR #2954, opened 2019-06-25
kernel: Move to Linux v5.1.15-openpower1

shenki:
Powerpc related changes:

 powerpc/mm/64s/hash: Reallocate context ids on fork
 powerpc/bpf: use unsigned division instruction for 64-bit operations
 powerpc/powernv: Return for invalid IMC domain
 powerpc/64s: Fix THP PMD collapse serialisation


shenki:
cc @mpe

stewart-ibm:
Contains the fix for CVE-2019-12817, which while important for distribution kernels, is not terribly important for ours in firmware, and not applicable for P9 at all for us (if you can change an NVRAM variable to change the petitboot kernel command line, then you have full access to the system anyway).

On P8 systems, you *could* use this to go from a locked down petituser to root in the petitboot shell, so anyone shipping a POWER8 system with a petitboot with password support should consider if a firmware update is appropriate.

shenki:
retest this please

mpe:
Yep, agree low impact for petitkernel. On affected systems it can also lead to soft lockups and other crashes, so you do still want it even without the security exposure.

stewart-ibm:
retest this please

stewart-ibm:
retest this please

PR #2980, opened 2019-07-05
configs/nicole: Add new platform configuration

artemsen:
The platform is a new storage controller for TATLIN server.
Based on IBM Romulus reference design (POWER9).


stewart-ibm:
The skiboot support will come in with https://github.com/open-power/op-build/pull/2981

stewart-ibm:
ok to test

wghoffa:
@artemsen -- This doesn't seem to work when I attempt to build the Nicole config. In the defconfig it points to a patch directory that doesn't exist. I have a fix here: https://github.com/open-power/op-build/pull/2986. Can you have a look?

PR #2986, opened 2019-07-09
Remove nicole config patches directory declaration

wghoffa:
- The definition of a Patches directory is pointing to something
    that does not currently exist so this config fails to build

Signed-off-by: Bill Hoffa <wghoffa@us.ibm.com>


wghoffa:
@artemsen -- Can you have a look? Without this change we get compilation errors that the patch directory it points to does not exist.

artemsen:
Thanks, William!
I used it during bring up and forgot to remove, sorry.

PR #2992, opened 2019-07-11
Add files for mihawk

chcyjoy:


stewart-ibm:
A couple of small things:
- please have the PR just have one commit that adds the new platform rather than also containing merge commits
- you will need to have the Developer Certificate of Origin (DCO), i.e. the commit needs to have your `Signed-off-by` line.

openpower/configs/hostboot/mihawk.config
```diff
@@ -0,0 +1,73 @@
+# The BMC MBOX Protocol is used to access PNOR
+unset SFC_IS_AST2500
+unset SFC_IS_AST2400
+set   PNORDD_IS_BMCMBOX
```
stewart-ibm: I'm pretty sure we want this being PNORDD_IS_IPMI. Is this right @amboar / @dcrowell77 ?

openpower/patches/mihawk-patches/occ/occ-20180525-fix-invalid-pwr-reading-when-no-raw-value.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/src/occ_405/amec/amec_sensors_power.c b/src/occ_405/amec/amec_sensors_power.c
+index bcd22c9..740cc08 100755
+--- a/src/occ_405/amec/amec_sensors_power.c
```
stewart-ibm: Is the OCC team happy to support this patch either forever or accept it upstream? @wilbryan ?

openpower/configs/mihawk_defconfig
```diff
@@ -0,0 +1,72 @@
...
+BR2_OPENPOWER_TARGETING_BIN_FILENAME="MIHAWK_HB.targeting.bin"
+BR2_OPENPOWER_TARGETING_ECC_FILENAME="MIHAWK_HB.targeting.bin.ecc"
+BR2_PACKAGE_PETITBOOT=y
+BR2_PACKAGE_PETITBOOT_MTD=y
+BR2_CAPP_UCODE_BIN_FILENAME="cappucode.bin"
+BR2_PACKAGE_LOADKEYS=y
+BR2_IMA_CATALOG_DTS="POWER9"
+BR2_OPENPOWER_SECUREBOOT_NO_KEY_TRANSITION=y
+BR2_OPENPOWER_SECUREBOOT_SIGN_MODE="development"
+BR2_BUILD_PNOR_SQUASHFS=y
```
stewart-ibm: mihawk is using vpnor?

openpower/patches/mihawk-patches/linux/Linux-nvme-pci-Fix-nvme-error-detected-kernel-crash.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/drivers/nvme/host/pci.c b/drivers/nvme/host/pci.c
+index 75539f7..2a96431 100644
+--- a/drivers/nvme/host/pci.c
++++ b/drivers/nvme/host/pci.c
```
stewart-ibm: Is this patch submitted upstream?

As per https://open-power.github.io/op-build/process/KernelTree.html we want all Kernel patches to be upstream.

In this case, the patch should also be in `git format-patch` format, so it has a full commit log, authorship details, and preferably where it was cherry-picked from.

openpower/configs/hostboot/mihawk.config
amboar: Yes. PNORDD_IS_BMCMBOX is deprecated. Do we have a good way of indicating that?

openpower/configs/mihawk_defconfig
amboar: @mine260309 should be able to answer this.

mine260309: Yes, it's decided to use whatever the same as Witherspoon, so it uses VPNOR.

openpower/patches/mihawk-patches/linux/Linux-nvme-pci-Fix-nvme-error-detected-kernel-crash.patch
HankChang: Hi @mikey , around ~03/2018, Wistron was hitting some petitboot kernel eeh/crash issues only when the nvme disks installed during their reboot regression testing.   At that moment, we got this simple patch from your side and the issue seems to be solved. :)   So, not sure if this patch is still needed or upstreamed already? or we could bypass/ignore this patch and then go ahead to test the "reboot+nvme" with current newer petitboot kernel? thank you.
# Related Boston issue/comments:
https://github.com/open-power/boston-openpower/issues/1076#issuecomment-372496320

openpower/patches/mihawk-patches/occ/occ-20180525-fix-invalid-pwr-reading-when-no-raw-value.patch
marthabroyles: Yes and there should be a request to upstream the change

openpower/configs/hostboot/mihawk.config
dcrowell77: Not really, we don't have a "openpower template" config file, though that is a good idea now that I say that out loud... ;-)

dcrowell77: And witherspoon.config still has PNORDD_IS_BMCMBOX set.  However, our code is actually written to allow either mode dynamically.  That was done to handle the non-crisp transition.  

So, the summary is - this is fine.

amboar: Yep, it's fine in that it will use try IPMI anyway, but it is a bit confusing. There's no reason not to switch all the P9 machines at this point. We shouldn't be adding new machines with a confusing config option

openpower/patches/mihawk-patches/linux/Linux-nvme-pci-Fix-nvme-error-detected-kernel-crash.patch
mikey: I just commented in that issue that this patch went upstream in v4.17 as:

> commit f0295e047fcf52ccb42561fb7de6942f5201b676
> Author: Michael Neuling <mikey@neuling.org>
> Date: Mon Mar 26 15:17:07 2018 +1100
> powerpc/eeh: Fix race with driver un/bind

So provided that does fix it, we should just be able to update and get the fix

openpower/configs/hostboot/mihawk.config
chcyjoy: We're follow the setting in Witherspoon and Zaius. Can we use PNORDD_IS_BMCMBOX currently? Until other P9 machines switch to IPMI, we'll follow them.

openpower/patches/mihawk-patches/occ/occ-20180525-fix-invalid-pwr-reading-when-no-raw-value.patch
chcyjoy: How should I send the request, using github pull request or any other ways?

chcyjoy: I sent the PR to occ.
https://github.com/open-power/occ/pull/27

chcyjoy:
Will create a new PR.

openpower/configs/hostboot/mihawk.config
wghoffa: Yes, the use of this is acceptable. Agree we should have the other P9 machines change first.

openpower/patches/mihawk-patches/linux/Linux-nvme-pci-Fix-nvme-error-detected-kernel-crash.patch
wghoffa: @HankChang -- Can we update to a kernel that includes the fix?

PR #3001, opened 2019-07-15
vesnin config: Add PNOR over HIOMAP support

artemsen:
Use IPMI/HIOMAP to work with PNOR flash.


stewart-ibm:
ok to test

PR #2994, opened 2019-07-11
buildroot: Move to 2019.05.1 stable release

shenki:
Upstream release notes:

 http://git.buildroot.net/buildroot/plain/CHANGES?id=2019.05.1
 http://lists.busybox.net/pipermail/buildroot/2019-July/254105.html

Note that this includes an upstreamed version of the localedef patch
that op-build was carrying.

Signed-off-by: Joel Stanley <joel@jms.id.au>

PR #3002, opened 2019-07-16
Kernel 5.2.1

shenki:


shenki:
This depends on https://github.com/open-power/op-build/pull/2994 going in

PR #3004, opened 2019-07-16
Remove POWER8 configurations

ghost:
We now only support POWER9 OpenPOWER platforms.

This is due to nobody being available to make the POWER8 hostboot tree
build with a GCC past gcc 6.x.


stewart-ibm:
Note that I'm submitting this now so that we have some time to sort out any CI problems because of this.

It shouldn't go in before we tag op-build v2.4 though.

danlarson-ibm:
IBM OP CI should go green. I've removed them from the matrix

stewart-ibm:
danlarson-ibm <notifications@github.com> writes:
> IBM OP CI should go green. I've removed them from the matrix

Okay. Is it just for op-build master builds?

we still have P8 systems in op-build 2.0.y, 2.3.y and will do in 2.4.y -
so they should be tested there

-- 
Stewart Smith
OPAL Architect, IBM.

danlarson-ibm:
Ah thats a good point. I didn't connect the dots on the other branches. I can address that.

stewart-ibm:
@danlarson-ibm looks like the P8 job still triggered, and unsurprisingly it failed :)

shenki:
I still use palmetto configurations for development and testing. Instead of removing all of them, I propose we declare them unsupported (stop doing internal CI and testing) but not remove from the tree.

danlarson-ibm:
I fixed up a filter for that. p8 will only run on op-build-* branches. 
I'm not sure how to get the check to be removed from here, once it has appeared

dcrowell77:
> I fixed up a filter for that. p8 will only run on op-build-* branches.

I think you will eventually have to be more specific here.  For example, v2.5 won't support p8.

> I still use palmetto configurations for development and testing. Instead of removing all of them, I propose we declare them unsupported (stop doing internal CI and testing) but not remove from the tree.

What's the best way to make that declaration?  Is there a way to throw messages out when you run the palmetto_defconfig?

oohal:
> I still use palmetto configurations for development and testing. Instead of removing all of them, I propose we declare them unsupported (stop doing internal CI and testing) but not remove from the tree.

Does your testing require the full firmware stack? We could trim down the P8 defconfigs to only build the kernel/initramfs if that's all you need.

danlarson-ibm:
> I think you will eventually have to be more specific here. For example, v2.5 won't support p8.
Done. op-build-\d\.[1-4][01]*\.y

stewartsmith:
Okay, so there seems to be enough community enthusiasm for at least some p8 platforms that we probably have to keep them around. Closing this PR.

PR #3006, opened 2019-07-16
Un-ship iprutils for OpenPower machines

shenki:
Petitboot has pulled in iprutils for the iprconfig tool but there is no
code in Petitboot that uses it. Furthermore, openpower systems do not
have this hardware so there is no need for it to be included.

Add it to the FSP based zz and firenze configurations as it may be used
there.


mikey:
@bjking1 you ok with this?

PR #3018, opened 2019-07-23
Remove all P8 platform except palmetto and vesnin

stewartsmith:
Palmetto and Vesnin are still used by people, all others are essentially
untested.

@shenki what about this as an alternative to https://github.com/open-power/op-build/pull/3004 ?


sharkcz:
Please keep habanero too, there was a big batch of them sold in Europe recently to various enthusiasts (like me). All built and worked OK ~month ago.

jcajka:
I second what @sharkcz wrote. I know several owners(incl. me) of [those](https://www.ebay.com/itm/Tyan-Rack-2U-Server-TN71-BP012-1x-IBM-POWER8-10-Core-CPU-2-926-GHz-2x-750Watt-/132754996458) and it is/was still the cheapest entry in to the OpenPower ecosystem for individuals(with over 64 sold). For me the op-build and upstream petitboot has been crucial as even with the latest vendor provided FW/petitboot I have not been able to boot any recent Fedora. It always resulted in crash in to the petitboot shell(I suspected grub cfg parsing issue). I'm currently running roughly ~2 months old build on it, without any noticeable issue.

stewartsmith:
> I second what @sharkcz wrote. I know several owners(incl. me) of [those](https://www.ebay.com/itm/Tyan-Rack-2U-Server-TN71-BP012-1x-IBM-POWER8-10-Core-CPU-2-926-GHz-2x-750Watt-/132754996458) and it is/was still the cheapest entry in to the OpenPower ecosystem for individuals(with over 64 sold). For me the op-build and upstream petitboot has been crucial as even with the latest vendor provided FW/petitboot I have not been able to boot any recent Fedora. It always resulted in crash in to the petitboot shell(I suspected grub cfg parsing issue). I'm currently running roughly ~2 months old upstream build on it, without any noticeable issue.

Ahhh. Hrm... I wonder what firmware Tyan shipped with them and what the actual problem is there... it's unfortunate if we ended up having something broken that would prevent booting.

So, in order to support running any test suite against Habanero and not going insane due to crazy-broken IPMI Serial Over LAN support I just submitted https://github.com/open-power/op-test/pull/531 which instantly found https://github.com/open-power/op-build/issues/3040 so already providing value :)

It looks like we do have a healthy enough community around Habanero and Palmetto that we really should keep them and evaluate how we're going to do that.

With luck, someone at IBM can get `op-test` going regularly on habanero and we can continue to have things tested.

Otherwise, if someone in the community wants to run some CI, I can certainly help with that.

jcajka:
> Ahhh. Hrm... I wonder what firmware Tyan shipped with them and what the actual problem is there... it's unfortunate if we ended up having something broken that would prevent booting.
> 

Last available image for it known to me is [here](https://tyan.com/Barebones=TN71-BP012=BSP012T71V14HR-4T-5(BTO)=downloads=EN). It is kind of old. I would suspect that Fedora is using some new feature in grub.cfg as for example Debian has been just fine(actually used it for the build and flash of the upstream version), but it could have been really anything. For the record I haven't seen any error messages just the crash in to the shell(IIRC prior to the petitboot "boot menu" showing), when there has been Fedora(30) install on any attached disk.

> So, in order to support running any test suite against Habanero and not going insane due to crazy-broken IPMI Serial Over LAN support I just submitted [open-power/op-test#531](https://github.com/open-power/op-test/pull/531) which instantly found #3040 so already providing value :)
> 
> It looks like we do have a healthy enough community around Habanero and Palmetto that we really should keep them and evaluate how we're going to do that.

Thank you very much :).

> 
> With luck, someone at IBM can get `op-test` going regularly on habanero and we can continue to have things tested.
> 
> Otherwise, if someone in the community wants to run some CI, I can certainly help with that.

What would be needed for that in general? 2 systems visible from internet?

dcrowell77:
> What would be needed for that in general? 2 systems visible from internet?

The systems under test wouldn't need to be internet visible, ours aren't.  Basically you would just need to plumb up a job through whatever infrastructure you are comfortable with (we use jenkins) that can see when new pull requests come into op-build and then triggers the tests.  At a first level without any special permissions/etc, you could just have the tests post a message to the pull request itself.  With a little more work things could probably get added to the more official 'scores' (that is outside my expertise).  It would be great if we were getting results from the platforms that we don't have, e.g. vesnin, nicole, etc

klauskiwi:
Reminder that there's a thread on the OpenPower-Firmware mailing-list discussing this: https://lists.ozlabs.org/pipermail/openpower-firmware/2020-July/000535.html

Based on the feedback so far (with the OpenPower foundation System-Software Workgroup, and the absence of it on the mailing-list), it's looking more like we're adopting the solution in #3145

klauskiwi:
#3145 includes this..

PR #3033, opened 2019-07-29
Ship PCI Utils

shenki:
This adds the PCIUTILS package back to the petitroot. In the past it was
included in the build as a dependency of IPRUTILS.

When that was dropped from the build we inadvertently dropped PCIUTILS,
which contains lspci. We have an lspci in busybox but as Oliver pointed
this out this is not adequate:

 > why did we replace the useful lspci in skiroot with the useless lspci?
 > the busybox lspci doesn't even show the domain number

Fixes: a1ae0bc353b8 ("Un-ship iprutils")


oohal:
Looks ok. I'm not sure what's up with the CI failure. Might be spurious?

stewartsmith:
For the DD2.3 test:

```
======================================================================
ERROR [8.209s]: runTest (testcases.SystemLogin.BMCLogin)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/hostboot-jenkins.swg-devops/jenkins-slave/workspace/Hostboot/OpenPower/HW/p9-test-param-new/op-test-framework/testcases/SystemLogin.py", line 112, in runTest
    self.assertTrue( len(found_issues) == 0, "REPORT_BUG BMC dmesg, debug log has full details:\n{}".format(msg))
AssertionError: REPORT_BUG BMC dmesg, debug log has full details:
[511842.158653] Killed process 1375 (mboxd) total-vm:303252kB, anon-rss:231516kB, file-rss:0kB, shmem-rss:0kB

----------------------------------------------------------------------```

which is totally unrelated to anything we could possibly do with this PR, so merging.

amboar:
FWIW that OOM [has been fixed](https://github.com/openbmc/hiomapd/commit/fac3689e77d37aca9c85dafabf921a847b1c4e02), just need newer BMC firmware on the CI nodes.

stewartsmith:
We might need to also do this for mihawk actually as that's been added recently, but merging this as it does fix a real problem now.

PR #3061, opened 2019-08-07
package/ocmb-explorer-fw: default to n

jk-ozlabs:
This package isn't currently used on any upstream OpenPOWER platforms,
so default to n. Platforms can enable it through their configs instead.


jk-ozlabs:
This allows us to build `legal-info` again.

PR #3084, opened 2019-08-15
Bump up version for openpower-pnor package

dcrowell77:
A backport of a fix in hostboot requires a bigger HBI partition so
moving up to a more recent level of the pnor package that includes
a bump.


dcrowell77:
retest this please

dcrowell77:
The DD2.3 fail is a known issue.  This PR is step1 in the path to fix it.

PR #3088, opened 2019-08-19
kernel: Move to Linux v5.2.9-openpower1

shenki:
PowerPC releated fixes:

 powerpc: fix off by one in max_zone_pfn initialization for ZONE_DMA


shenki:
The failure was due to a BMC bug. As the other tests have passed, we have good coverage of the new kernel building and booting, so I will merge this.

PR #3094, opened 2019-08-20
Update openpower-pnor makefile to use BR2_ version vars

crgeddes:
Previously we were using the * wildcard to lookup files to copy
over for debug data. This breaks when there are multiple versions
of a given package built in the output/build directory. This commit
will make the openpower-pnor.mk files use the build root env vars
(BR2_..) to make sure we just attempt to copy files from the current
version we are using.


dcrowell77:
@klauskiwi

oohal:
looks good to me

PR #3118, opened 2019-09-04
Documentation: Update Readme.md for Fedora installs

klauskiwi:
A default Fedora29 ppc64el install is missing a few dependent packages in order to be able to build op-build


klauskiwi:
Even with that change, seems like op-build is failing to build on ppc64el Fedora29, specifically when building HCODE... investigating...

PR #3123, opened 2019-09-06
Update openpower-pnor package

crgeddes:
This will revert a commit that was causing swift compiles to fail.

Signed-off-by: Christian Geddes crgeddes@us.ibm.com

### DIFF
--- a/openpower/package/openpower-pnor/openpower-pnor.mk
+++ b/openpower/package/openpower-pnor/openpower-pnor.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-OPENPOWER_PNOR_VERSION ?= e07c9e5aa4bdb96ea02a7fc30bf7c67ae7ebeea5
+OPENPOWER_PNOR_VERSION ?= 82861b1c703839098b1800b228493f775009e6c8
 OPENPOWER_PNOR_SITE ?= $(call github,open-power,pnor,$(OPENPOWER_PNOR_VERSION))
 
 OPENPOWER_PNOR_LICENSE = Apache-2.0


stewartsmith:
Swift isn’t upstream, so unless this causes a regression for an upstream platform, I do not see why this should go in.

In the past we have not provided any guarantees for out of tree code.

crgeddes:
This change is backing out changes that were made to axonePnorLayout64.xml (publicly available) which is only used in the swift config. IBM is running CI's internally that are pointing at the XML in order to keep regressions on swift running. 

Without this we would have to make wait for the real fix, meanwhile our CI's are failing and we cannot get anything through. 

Unless you can show how this hurts anyone, I would really like to get this in. Arguably this is a commit that should have never been merged, hence why its a revert.

crgeddes:
Basically this is reverting our attempts to get swift upstream because it broke things for us. No upstream users will be affected.

stewartsmith:
> On 6 Sep 2019, at 12:01, Christian Geddes <notifications@github.com> wrote:
> 
> This change is backing out changes that were made to axonePnorLayout64.xml (publicly available) which is only used in the swift config. IBM is running CI's internally that are pointing at the XML in order to keep regressions on swift running.
> 
> Without this we would have to make wait for the real fix, meanwhile our CI's are failing and we cannot get anything through.
> 
> Unless you can show how this hurts anyone, I would really like to get this in. Arguably this is a commit that should have never been merged, hence why its a revert.
> 

Makes sense now, thanks. Probably a bit better to revert the commit in pnor repo though

crgeddes:
These changes were picked up in another PR so I am closing this.

PR #3145, opened 2019-09-13
Remove support for POWER8

bofferdn:
Remove configurations and packages related to POWER8 machines going forward

Signed-off-by: Nick Bofferding <opensource@bofferding.net>


oohal:
Same comments as https://github.com/open-power/op-build/pull/3018 apply?

klauskiwi:
Reminder that there's a thread on the OpenPower-Firmware mailing-list discussing this: https://lists.ozlabs.org/pipermail/openpower-firmware/2020-July/000535.html

Based on the feedback so far (with the OpenPower foundation System-Software Workgroup, and the absence of it on the mailing-list), it's looking like we'll revamp this patch and apply it soon.

klauskiwi:
Considering the feedback we had so far, I'm re-basing this on top of the latest master so that we can review and merge it.. I expect some CIs will fail, as they'll be attempting to build P8 targets

klauskiwi:
For those not able to inspect the CI results above, we successfully built the usual Power9 targets, and also successfully executed the usual op-test cycle on top of P9 platforms. With that, I believe this is ready to be merged...

PR #3162, opened 2019-09-23
Add support for extended PPE42 instructions

dgilbert999:


dgilbert999:
This failed due to this error:
ERROR: ../img/hbibl.bin too large. Max allowed size is 32768
It does not seem to be related to this change.

dcrowell77:
yes, that is a *different* bug that we only see on ppc compiles...  changes are coming to fix that as well

klauskiwi:
I included the changes here in PR #3260 - closing

PR #3167, opened 2019-09-25
buildroot: Update to pick up new linux-firmware

shenki:
The 5.2 kernel requires version 7.13.11.0 of the Broadcom bnx2x
firmware. This can be found in the latest linux-firmware package.

A corresponding change has been made to the firmware whitelist script so
the new version is included.

This fixes the following symptoms:

 bnx2x 0003:01:00.0: Direct firmware load for bnx2x/bnx2x-e2-7.13.11.0.fw failed with error -2

### DIFF
--- a/openpower/scripts/firmware-whitelist
+++ b/openpower/scripts/firmware-whitelist
@@ -6,7 +6,7 @@
 # slash is required.
 whitelist=(     'acenic/'
                 'bnx2/'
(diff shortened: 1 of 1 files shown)


oohal:
@klauskiwi ping

klauskiwi:
> @klauskiwi ping

Yep, this looks alright

PR #3168, opened 2019-09-25
kernel: Move to Linux v5.2.17-openpower1

shenki:
PowerPC fixes:

 powerpc/mm/radix: Use the right page size for vmemmap mapping
 powerpc: Add barrier_nospec to raw_copy_in_user()
 powerpc/tm: Fix restoring FP/VMX facility incorrectly on interrupts
 powerpc/tm: Fix FP/VMX unavailable exceptions inside a transaction


klauskiwi:
Thanks Joel. Don't see a reason not to merge.

PR #3180, opened 2019-10-01
kernel: Move to Linux v5.3.1-openpower1

shenki:


openpower/linux/0001-xhci-Reset-controller-on-xhci-shutdown.patch
```diff
@@ -14,7 +14,7 @@ Signed-off-by: Joel Stanley <joel@jms.id.au>
  1 file changed, 1 insertion(+)
 
 diff --git a/drivers/usb/host/xhci.c b/drivers/usb/host/xhci.c
-index 3f79f35d0b19..e0f5518e7f2f 100644
+index 03d1e552769b..1e1624f5be9f 100644
```
klauskiwi: Just for my own education - do you use a script or any other tricks to re-base and update the .patch files?

shenki: We put all of the patches in the linux tree, as described here: https://open-power.github.io/op-build/process/KernelTree.html

When I bump a kernel release we use git-format-patch to create a patch series, and place them in op-build/openpower/linux.

shenki:
> Probably nitpicking, but would you be able to also update doc/process/KernelTree.rst in the same PR? Otherwise I can push it later, so approving

Not a nit pick at all, that's a good suggestion. I have made that change and will merge.

PR #3189, opened 2019-10-03
skiboot v6.5-27-g1785745d

opalci:
Jordan Niethe (1):
      ccan: Add CCAN heap license

Madhavan Srinivasan (1):
      hdat/spira: export abc and wxyz bus speeds

Nicholas Piggin (6):
      core/cpu.c: current_radix_mode should be initialised to true
      build: use -mcpu=power8
      core/exceptions.c: rearrange code to allow more interrupt types
      core/util: trap based assertions
      core/util: branch-to-NULL assert for ELFv2 ABI
      core/exceptions.c: do not include handler code in exception backtrace

Oliver O'Halloran (1):
      core/platform: Actually disable fast-reboot on P8


hegdevasant:
ok to test

klauskiwi:
This looks great, but the only test being performed is the defconfig sanity check. @op-jenkins @hegdevasant how can we enable the P8 and P9 IPL tests to this changes, same as we have been doing for Hostboot changes?

hegdevasant:
> This looks great, but the only test being performed is the defconfig sanity check. @op-jenkins @hegdevasant how can we enable the P8 and P9 IPL tests to this changes, same as we have been doing for Hostboot changes?

@klauskiwi IIUC CI gets triggered if PR is raised by open-power members. You may have to add `opalci` user to `open-power` member list. Checking with @oohal .

-Vasant

hegdevasant:
ok to test

dcrowell77:
We have reason to believe this change is breaking all of our P9 OP systems at the moment.  It looks like the CI tests never actually ran against this change.

PR #3193, opened 2019-10-04
buildroot: Update to 2019.05.3

shenki:
Upstream release notes:

  http://lists.busybox.net/pipermail/buildroot/2019-October/261556.html
  https://git.buildroot.net/buildroot/plain/CHANGES?id=2019.05.3


shenki:
2019.08 drops support for GCC 6, which our (P8) hostboot builds require.

PR #3223, opened 2019-10-21
Fix issue that PnorUtils.pm places new layout file inside of buildroot

LF-Luis:
PnorUtils.pm’s loadPnorLayout now accepts an optional argument specifying where to place the new layout file. If argument is not given, it will be placed in the directory where the script is running.

Signed-off-by: Luis Fernandez <Luis.Fernandez@ibm.com>

### DIFF
--- a/openpower/package/hostboot/0001-loadPnorLayout-accepts-optional-parameter.patch
+++ b/openpower/package/hostboot/0001-loadPnorLayout-accepts-optional-parameter.patch
@@ -1,0 +1,208 @@
+From c4100b7113a6ec9b70678f683107e7700a8a42da Mon Sep 17 00:00:00 2001
+From: Luis Fernandez <Luis.Fernandez@ibm.com>
+Date: Fri, 18 Oct 2019 15:53:23 -0500
+Subject: [PATCH] loadPnorLayout accepts optional parameter to place edited xml
+ layout file.
+
+Change-Id: I7b9ee0f7ad33ad88be58598ddf594765eabe29ea
+---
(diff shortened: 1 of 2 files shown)


stewartsmith:
are these patches headed upstream? is there a PR to track them?

openpower/package/hostboot/0001-loadPnorLayout-accepts-optional-parameter-to-place.patch
```diff
@@ -0,0 +1,115 @@
...
++
+     my $filename = basename($i_pnorFile, ".xml");
+-    $filename = "${filename}WithOffsets.xml";
++
++    if (defined $i_outputLayoutLocation)
++    {
++        $filename = "${i_outputLayoutLocation}/${filename}WithOffsets.xml";
++    }
++    else {
++        # If argument $i_outputLayoutLocation is not supplied, the new layout
```
artemsen: I think it is better to do not create the file at all, if output location is not specified.

lpfluisfer:
> are these patches headed upstream? is there a PR to track them?

@stewartsmith I placed both patches here in an op-build commit to test in CI and get some feedback. Ultimately I'll make separate commits to the Hostboot and PNOR repo.

openpower/package/hostboot/0001-loadPnorLayout-accepts-optional-parameter-to-place.patch
lpfluisfer: Agree, made changes to reflect that.

lpfluisfer:
The PNOR patch as a PR:
https://github.com/open-power/pnor/pull/130

PR #3240, opened 2019-10-25
Enable Radeon driver on Petitboot

klauskiwi:
There are use-cases requiring the Radeon DRM driver to be enabled, so
that the Petitboot console output to be seen through a display connected
to it at boot. The radeon drivers were disabled until recently due to a
driver bug preventing it to work properly with kexec, which have been
fixed as of kernel 5.3.7


klauskiwi:
@shenki @oohal please review... at this point I'm not sure if we're hitting expected failures from the CI or we're really beyond the max size for the pnor

```
cd build/buildpnor/ && /usr/bin/make --no-print-directory --no-builtin-rules --no-builtin-variables buildpnor
make[6]: Nothing to be done for 'buildpnor'.
true
../src/build/trace/tracehash_hb.pl -c -d ../obj -s ../img/hbotStringFile
    GENLIST    hbibl
ERROR: ../img/hbibl.bin too large. Max allowed size is 32768
makefile:435: recipe for target 'imgsizecheck' failed
make[5]: *** [imgsizecheck] Error 1
    GENLIST    hbicore
make[5]: *** Waiting for unfinished jobs....
    GENLIST    securerom
    MAKE       bootloader IMAGE
    GENLIST    hbicore_test
    GENLIST    hbirt_test
    GENLIST    hbirt

Combining Hash strings...
```

klauskiwi:
retest this please

oohal:
> Given the radeon driver is broken when kexecing, I don't think we are ready to enable the radeon driver in petitboot: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8d13c187c42e110625d60094668a8f778c092879
> 
> Users who wish to use petitboot on their monitors should use the aspeed graphics.

Please do not encourage VGA using. That's just mean.

Do we need to carry that patch in the petitkernel? We've carried the xHCI kexec patch since forever so it doesn't seem like a major ask.

klauskiwi:
> 
> 
> > Given the radeon driver is broken when kexecing, I don't think we are ready to enable the radeon driver in petitboot: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8d13c187c42e110625d60094668a8f778c092879
> > Users who wish to use petitboot on their monitors should use the aspeed graphics.
> 
> Please do not encourage VGA using. That's just mean.
> 
> Do we need to carry that patch in the petitkernel? We've carried the xHCI kexec patch since forever so it doesn't seem like a major ask.

I agree that, considering sound use case / justification and the fact that this runs fine on Power Systems, we could probably carry this while it's being sorted out upstream (a bug is a bug, and a partial solution is one type of solution).

At this point I can't figure out, though, why the CI failed, and the logs are gone, so I'll ask to retest this please.

klauskiwi:
So looks like the resulting image is too big. From the http://blade4a.ozlabs.ibm.com:8080/job/op-build/job/op-build-PR/1288/ logs:

```
<...>
SECBOOT section size: 147456, bytes used: 147456, bytes unused: 0
ATTR_PERM section size: 32768, bytes used: 32256, bytes unused: 512
NVRAM section size: 589824, bytes used: 589824, bytes unused: 0
BOOTKERNEL section size: 16252928, bytes used: 32505856, bytes unused: -16252928
ERROR: PnorUtils::checkSpaceConstraints: Image provided (/scratch/jenkins/workspace/op-build/op-build-PR/PLATFORM/sdk/label_exp/austin/output-ubuntu1804-witherspoon_defconfig/host/powerpc64le-buildroot-linux-gnu/sysroot/openpower_pnor_scratch//zImage.epapr) has size (32505856) which is greater than allocated space (16252928) for section=BOOTKERNEL.  Aborting! at /scratch/jenkins/workspace/op-build/op-build-PR/PLATFORM/sdk/label_exp/austin/output-ubuntu1804-witherspoon_defconfig/host/powerpc64le-buildroot-linux-gnu/sysroot/hostboot_build_images/PnorUtils.pm line 446.
<...>

@oohal any thoughts on how to proceed?

```

KyleMahlkuch:
The original radeon patch breaks some x86 boxes so I'm still working on getting that in the upstream kernel. 
There's a chance some of these radeon firmware packages aren't needed in petitboot. We don't need 3D graphics, etc in petitboot so that could potentially make the image smaller if we can determine which are needed / not needed. I will keep looking into this option.

klauskiwi:
Closing while this is better established and justified

PR #3246, opened 2019-10-25
Modify mihawk config files

chcyjoy:
Modify mihawk config to sync up settings from witherspoon.

Signed-off-by: Joy Chu <joy_chu@wistron.com>


oohal:
retest this please

PR #3247, opened 2019-10-25
libflash: skiboot v6.5-61-g66ab3cbd

opalci:
We have merged 100+ pathes after v6.4. I didn't realize that we have
separate package for libflash and missed to update :-(


opalci:
Note that for now I have created separate PR for libflash. Eventually I will integrate this with skiboot PR and will have one PR.

-Vasant

PR #3283, opened 2019-11-08
Modify kernel partition size for some built-in tools

chcyjoy:
Add openpower-pnor patch file to increase Bootloader Kernel partition size to 20MB for built-in RAID tool support

Signed-off-by: Joy Chu <joy_chu@wistron.com>

### DIFF
--- a/openpower/configs/mihawk_defconfig
+++ b/openpower/configs/mihawk_defconfig
@@ -1,5 +1,6 @@
 BR2_powerpc64le=y
 BR2_powerpc_power8=y
+BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_OP_BUILD_PATH)/patches/mihawk-patches"
 BR2_BINUTILS_EXTRA_CONFIG_OPTIONS="--enable-targets=powerpc64-linux"
 BR2_GCC_VERSION_6_X=y
 BR2_EXTRA_GCC_CONFIG_OPTIONS="--enable-targets=powerpc64-linux --disable-libsanitizer"
--- a/openpower/patches/mihawk-patches/openpower-pnor/openpower-pnor-20191118-change-kernel-size-to-24MB.patch
+++ b/openpower/patches/mihawk-patches/openpower-pnor/openpower-pnor-20191118-change-kernel-size-to-24MB.patch
@@ -1,0 +1,161 @@
+diff --git a/p9Layouts/defaultPnorLayout_64.xml b/p9Layouts/defaultPnorLayout_64.xml
+index 3f74223..86b8053 100644
+--- a/p9Layouts/defaultPnorLayout_64.xml
++++ b/p9Layouts/defaultPnorLayout_64.xml
+@@ -222,10 +222,10 @@ Layout Description
+         <readOnly/>
+     </section>
+     <section>
+-        <description>Bootloader Kernel (15.5MB)</description>
++        <description>Bootloader Kernel (24MB)</description>
+         <eyeCatch>BOOTKERNEL</eyeCatch>
+         <physicalOffset>0x2141000</physicalOffset>
+-        <physicalRegionSize>0xF80000</physicalRegionSize>
++        <physicalRegionSize>0x1800000</physicalRegionSize>
+         <side>A</side>
+         <sha512Version/>
+         <readOnly/>
+@@ -233,7 +233,7 @@ Layout Description
+     <section>
+         <description>OCC Lid (1.125M)</description>
+         <eyeCatch>OCC</eyeCatch>
+-        <physicalOffset>0x30C1000</physicalOffset>
++        <physicalOffset>0x3941000</physicalOffset>
+         <physicalRegionSize>0x120000</physicalRegionSize>
+         <side>A</side>
+         <sha512Version/>
+@@ -243,7 +243,7 @@ Layout Description
+     <section>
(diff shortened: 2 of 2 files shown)


stewartsmith:
A better way to do this would be to have a partition for petitboot plugins.

Bumping the size of the initramfs slows down boot. It's about 1MB/sec to read from flash over LPC, and 16->20MB means 4 extra seconds of boot time.

chcyjoy:
@stewartsmith We take p9dsu's commit as a reference.
https://github.com/supermicro/p9dsu-op-build/blob/master/openpower/custom/patches/openpower-pnor/0001-bootloader-kernel-size-to-24-MB.patch

@jk-ozlabs There isn't any partition for petitboot. What do you think about it?

--joy chu

jk-ozlabs:
Like @stewartsmith has mentioned, it'd be much better as a separate partition for those built-in tools, especially with the different licenses in use there (I assume? there doesn't seem much info about those here). That would be a matter of adding a new partition, which petitboot would discover as a source for plugins; a lot of the infrastructure already exists for that.

Additionally, the extra `BOOTKERNEL` size will slow down boot, and it's one of the larger contributors to boot time.

That said, it's specific to wistron machines, so if those concerns are acceptable, it's really your call to make.

chcyjoy:
@stewartsmith After our evaluation, we decide to change the kernel partition size to 24MB. The increase of the boot time is acceptable for us.

Due to the tight schedule, would you help us merge the commit? We'll remove this patch if there's code base update upstream or a new petitboot partition in the future.

--joy chu

klauskiwi:
> @stewartsmith After our evaluation, we decide to change the kernel partition size to 24MB. The increase of the boot time is acceptable for us.
> 
> Due to the tight schedule, would you help us merge the commit? We'll remove this patch if there's code base update upstream or a new petitboot partition in the future.
> 
> --joy chu

Joy,

Considering this change is limited to Mihawk only, and that apparently we don't have the time/resources to do anything better (with another partition as suggested), I believe we can merge this patch - However, there's no way I can tell it works correctly, since there's no CI or test results posted here or in OP940.

Please provide some evidence that this was tested in Mihawk and please plan to add Mihawks to upstream (here) and downstream CI as well..

chcyjoy:
> > @stewartsmith After our evaluation, we decide to change the kernel partition size to 24MB. The increase of the boot time is acceptable for us.
> > Due to the tight schedule, would you help us merge the commit? We'll remove this patch if there's code base update upstream or a new petitboot partition in the future.
> > --joy chu
> 
> Joy,
> 
> Considering this change is limited to Mihawk only, and that apparently we don't have the time/resources to do anything better (with another partition as suggested), I believe we can merge this patch - However, there's no way I can tell it works correctly, since there's no CI or test results posted here or in OP940.
> 
> Please provide some evidence that this was tested in Mihawk and please plan to add Mihawks to upstream (here) and downstream CI as well..


After adding the patch to increase kernel partition, we put the tools under openpower/overlay/bin. The PNOR can be built successfully.

The pnor version is
### open-power-mihawk-v20191120_kernel24MB
        op-build-v2.4-246-g74277a39
        buildroot-2019.05.3-12-gd2eb9d62c1
        skiboot-v6.5-61-g66ab3cbd
        hostboot-9c7a264
        occ-3ab2921
        linux-5.3.7-openpower1-p0be1ca5
        petitboot-v1.10.4
        machine-xml-f6ec5aa
        hostboot-binaries-hw080119a.940
        capp-ucode-p9-dd2-v4
        sbe-2742319
        hcode-hw110519a.opmst

 The op-build log are as below.
[mihawk_opbulid_201911201402.zip](https://github.com/open-power/op-build/files/3868942/mihawk_opbulid_201911201402.zip)
[mihawk_rebuild_kernel_pnor_201911201724.log](https://github.com/open-power/op-build/files/3868920/mihawk_rebuild_kernel_pnor_201911201724.log)

We have two tools, storcli64 and sas3flash, now. Here's the log that i tried some commands under petitboot.
[MobaXterm_10.32.49.2p2200BMC_20191120_191107_storcli_sas3flash.txt](https://github.com/open-power/op-build/files/3868933/MobaXterm_10.32.49.2p2200BMC_20191120_191107_storcli_sas3flash.txt)

klauskiwi:
@crgeddes @lpfluisfer Christian and Luis, this looks correct and safe to me, but I'm not completely comfortable with the PNOR layout file. Can you review and check if this can be merged (note this is patching the PNOR layout for Mihawk only).

Also let us know if you prefer this handled differently (i.e., a separate pnor layout file for mihawk, or something else)

youhour:
can someone tell me whether the patch just to increase the partition size or it also include the tool binary images?   Thanks in advance.

crgeddes:
@youhour The commits in this pull-request are only increasing the size of the BOOTKERNEL section of the mihawk pnor layout. From what I see there is no mention of what is going to be loaded into the newly allocated space.

@klauskiwi I am okay with these changes. Math looks right and the total size is still under 64 MB

youhour:
@crgeddes Thank you Christian.  That was what I was hopping for.

stewartsmith:
This is the same excuse used previously. It really should be done properly rather than a continuous string of hacks.

Using a patch is a rather fragile mechanism for this too, and with no in tree usage of the extra space, any rework to flash assembly (see my patches on this) is likely to just make every platform the same.

PR #3318, opened 2019-12-03
Add hostboot submodule update

strange671:
Hostboot now includes libmctp, which is imported as a git submodule. Thus, in order to build Hostboot with MCTP (Management Component Transport Protocol) support, we must clone the git tree and its sub-modules.

Signed-off-by: Jet Li <Jet.Li@ibm.com>


stewartsmith:
I don’t understand the reasoning from the text in the PR. Can you elaborate?

strange671:
In Hostboot that has submodule config, but it will not also download while to build. So I want to make a global change when submodule exists that it will clone.

stewartsmith:
That doesn't make anything clearer.

From peering into the Hostboot repo, how about this:


"Hostboot now includes libmctp, which is imported as a git submodule. Thus, in order to build Hostboot with MCTP (Management Component Transport Protocol) support, we must clone the git tree and its sub-modules."

-- 
  Stewart Smith
  stewart@flamingspork.com

bofferdn:
The git submodule makefile directive won't work here since it requires the git site method, which is not configured for this package.

strange671:
Many thanks, I've edited it. For site_method that does not git, I'll drop this PR.

PR #3340, opened 2019-12-15
Linux v5.4.3-openpower1

stewartsmith:
All the out of tree patches are merged. We have to keep v5.3 headers or wait for buildroot to bump.

For me, needed for fixing NVME bugs (which could have been v5.3.8, but may as well go to v5.4).

It compiles, about to boot test :)


oohal:
If we're using 5.4 are the BR config files supposed to have BR2_KERNEL_HEADERS_5_3=y everywhere?

stewartsmith:
> The way we've done this in the past is to bump to use the same kernel headers for the toolchain/userspace as we use for the kernel.

Yeah, just have to bump buildroot to do that.

shenki:
I've put this up which will hopefully help you out with your nvme issues in the short term

 https://github.com/open-power/op-build/pull/3342

When someone finds time to bump buildroot, we can move to the 5.4 kernel.

shenki:
Thanks for the patch. We've now moved to 5.4.7, so I'll close this.

stewartsmith:
Perfect, thanks.

PR #3341, opened 2019-12-15
RaptorCS Blackbird support

stewartsmith:
Hostboot config is basically Witherspoon but with LPC PORT 80h turned on
defconfig is from Witherspoon, with Blackbird XML setup from Raptor's tree.

NOTE: without linux v5.4.3 (or at least anything more recent than v5.3.7) my Intel SSD doesn't show up in Petitboot.

Also note that the blackbird-xml project doesn't yet existing in the open-power org, so you need to grab it from my repo (or Raptor's). Consider this a request to branch mine in :)

Signed-off-by: Stewart Smith <stewart@flamingspork.com>


stewartsmith:
I've sent the skiboot patch upstream: https://patchwork.ozlabs.org/patch/1208967/

sharkcz:
Hmm, I guess I should make the same for Talos :-)

madscientist159:
Did the IPL observer stuff ever make its way in to upstream?  Especially w/ Talos II it's pretty critical. ;)

stewartsmith:
On Sun, Dec 15, 2019, at 10:45 PM, Timothy Pearson wrote:
> Did the IPL observer stuff ever make its way in to upstream? Especially 
> w/ Talos II it's pretty critical. ;)

Not sure on the BMC side.

The host side, we have it upstream in skiboot, but the hostboot part isn't there, and there's nothing in the petitboot environment which would do the final write of 254,254.

sharkcz:
I have started https://wiki.raptorcs.com/wiki/Firmware_Upstreaming (it got few changes from @merklort for the hostboot part) some time ago, probably based on the Talos GA branch (2018-04-19), not on the v2 firmware (2019-04-16).

madscientist159:
I know that was one of the main blockers to upstreaming -- we don't want to have the support issues associated with OCC reads during IPL (or completely disables fan controls, for that matter).  The former in particular was quite expensive as the symptoms lead to spurious mainboard / CPU RMA in some cases and escalation to higher level support in many others.

While I'd really like to see upstream support, we need to have the IPL observer support merged into the subcomponents first.  Any chance that can happen?

sharkcz:
Can't we start with the machine specific "overlay" in openpower/patches/talos-patches until they are properly upstreamed? How much of the non-upstream patches is shared between Blackbird and Talos?

stewartsmith:
It’s probably a fair bit. I’ll try and get an updated PR sometime in the next day or so

dcrowell77:
There were some internal network issues recently so I'm guessing that is what the CI fail probably is.

dcrowell77:
retest this please

madscientist159:
> Can't we start with the machine specific "overlay" in openpower/patches/talos-patches until they are properly upstreamed? How much of the non-upstream patches is shared between Blackbird and Talos?

Yes, a local set of patches combined with a RCS overlay directory would work.  Most of the modifications are common between Talos II / Blackbird / Condor; please feel free to pull whatever you need from our repos as we have retained the original Apache / GPL / BSD licenses for the modified components (newly written components are normally GPL v3 from our side, licenses are in the file headers).

stewartsmith:
The v2 for skiboot patch, which talks IPL Observer: https://patchwork.ozlabs.org/patch/1210958/

oohal:
retest this please

madscientist159:
@stewartsmith Getting the IPL observer stuff into the SBE and hostboot would also be good -- in particular, the SBE patches (in our tree) drive the progress bar during the first part of the Blackbird boot, and have proven very useful over time to determine bad CPUs / sockets from hostboot issues at a glance.

stewartsmith:
hoping we can have someone still working for IBM push the SBE patches, I’m a bit less inclined to mess with mine now that I don’t have crorecoversbe laying around...

madscientist159:
I'd be willing to test any patches to upstream, since I do have access to the recovery system...

madscientist159:
Just dug this back up:
https://github.com/open-power/sbe/pull/15

sharkcz:
I have updated https://wiki.raptorcs.com/wiki/Firmware_Upstreaming with SBE, HCODE, OCC and Hostboot. Seems there is just few changes in the Raptor branches for these components.

madscientist159:
Most center around the IPL observer, bit frustrating to have the simple PRs sit for > 8 months TBH.

dcrowell77:
Is the SBE PR the only one outstanding?  I understand the frustration, I'll try to shake a tree.

sharkcz:
If I see right in the wiki page above, then we need https://git.raptorcs.com/git/talos-hostboot/commit/?id=d90e6c513094231f622a427030f3dbca1eeb5ed5 for the hostboot part. @madscientist159 will know if there has been a PR already.

dcrowell77:
Of course the link was added after I read it.  It looks like 3 new commits.  Make some PRs and I'll at least get them into the pipeline (with no promises on getting a lot of attention).

sharkcz:
op-build WIP with Talos support = https://github.com/sharkcz/op-build/tree/talos
skiboot WIP with Talos using LPC observer = https://github.com/sharkcz/skiboot/tree/talos-lpc

be aware, not even compile tested

madscientist159:
FWIW 0xfefe is the IPL observer code for Linux online and userspace launched.  Skiboot should NOT be sending it; it confuses two parts of the process and tells the BMC IPL has successfully finished when in fact it's only partially done.

sharkcz:
@madscientist159, I understand your position. The open question is, whether there is a way to do the signalling correctly and upstreamable. How do the other platforms handle this kind of communication between the host and BMC?

madscientist159:
@sharkcz I've seen a bunch of different methods over the years including writes to scratch registers in the south bridge.  What is wrong with a small userspace tool poking the last status code out the LPC port?

sharkcz:
@madscientist159, nothing is wrong, it's for my education and overview :-) Also I'm thinking how it could be integrated into upstream.

madscientist159:
@sharkcz we actually need an LPC access module for other reasons, something that can allow root to read / write bytes, nothing fancy.  I can see why it might be desirable to disable the debug interface, but LPC access is a useful and reasonable thing to have.

I wonder if we can do a simple device node that takes basic commands like IO_LPC_SET_TARGET_ADDRESS, IO_LPC_WRITE_BYTE, IO_LPC_READ_BYTE?

stewartsmith:
We won’t have debugfs mounted in any secure mode

stewartsmith:
What alternate code can I send down that kicks on fan control? By the time skiboot exits there’s nothing touching anything that wouldn’t happen during Normal runtime

madscientist159:
Would there be an objection to an LPC access module in secure mode, if we were to write one?

stewartsmith:
Something that did just this poke? Probably not. But arbitrary access? Yeah.

madscientist159:
Yeah was looking at arbitrary access, see my post a few above about a possible IO_ access mechanism around /dev/lpc or similar.

PR #3369, opened 2020-01-02
buildroot: Support kernel 5.4

shenki:
This picks up a buildroot commit to support Linux 5.4. It does not
change the default.


shenki:
Superseded by https://github.com/open-power/op-build/pull/3373

PR #3408, opened 2020-01-21
kernel: Enable security options

shenki:


shenki:
This change depends on userspace support for kexec_file_load. See https://github.com/open-power/op-build/issues/3418

shenki:
The patch no longer sets FORCE_INTEGRITY=y, so it can be merged independnetly of userspace support for kexec_file.

PR #3443, opened 2020-02-11
P9 DTS file for separate core/thread events and NVLINK fix

kgdeepak:
P9 DTS file for separate core/thread events and NVLINK fix

### DIFF
--- a/openpower/package/ima-catalog/ima-catalog.mk
+++ b/openpower/package/ima-catalog/ima-catalog.mk
@@ -3,7 +3,7 @@
 # ima-catalog.mk
 #
 ################################################################################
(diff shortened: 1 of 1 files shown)


oohal:
ok to test

oohal:
For future reference, when you send a PR to op-build updating a package it should only have a single commit that changes the package SHA. This history isn't interesting to us and it's a potential bisect hazard.

Squashed and merged.

kgdeepak:
Thank you. There was some issue while trying to sync my fork and while pushing to op-build master. Hence this problem came up. In future, i will ensure what you suggested.

PR #3451, opened 2020-02-15
libflash.mk: fix version and add dependencies

klauskiwi:
LIBFLASH_VERSION is not set correctly when using a
BR2_SKIBOOT_CUSTOM_GIT, causing the package to
not be downloaded.

Additionally, add skiboot as a build-time dependency

### DIFF
--- a/openpower/package/libflash/libflash.mk
+++ b/openpower/package/libflash/libflash.mk
@@ -4,13 +4,16 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


hegdevasant:
Looks like s-o-b is missing? DCO verification is not passing.

Otherwise patch looks fine to me.

-Vasant

klauskiwi:
I guess I was out of coffee, tried doing this trivial fix on Github itself and ended up creating a branch.. I created PR #3452, will close this one and sort things out

PR #3454, opened 2020-02-16
kernel: Move to Linux v5.4.18-openpower1

runlevel5:
Upstream powerpc related changes:

  * powerpc/fsl/dts: add fsl,erratum-a011043

Signed-off-by: Trung Lê <trung.le@ruby-journal.com>


openpower/linux/0003-Release-OpenPower-kernel.patch
```diff
@@ -15,7 +15,7 @@ index e16d2e58ed4b..cabff177cbb4 100644
 @@ -2,7 +2,7 @@
```
klauskiwi: Probably nitpicking, but the lack of changes in the index line indicates that you changed the patch itself, and have not re-applied/git-format-patch'ed.. perhaps you should do that?

Otherwise, 18 configs checked-in, minor-version kernel bump, LGTM

oohal:
ok to test

PR #3471, opened 2020-02-25
Update Blackbird support

stewartsmith:
Pull in Hostboot port 81h/82h progress code
Update Skiboot to report end of IPL
Make blackbird patches Blackbird specific

Signed-off-by: Stewart Smith <stewart@flamingspork.com>

### DIFF
--- a/openpower/patches/blackbird-patches/skiboot/0001-platform-add-Raptor-Blackbird-support.patch
+++ b/openpower/patches/blackbird-patches/skiboot/0001-platform-add-Raptor-Blackbird-support.patch
@@ -1,4 +1,4 @@
-From 376bed3fbe902138b92e95d9da58bb5e1509072e Mon Sep 17 00:00:00 2001
+From 2a4f8f6f67d1e0748bd20b4f721607ce7855ad30 Mon Sep 17 00:00:00 2001
 From: Stewart Smith <stewart@flamingspork.com>
 Date: Sun, 8 Dec 2019 19:06:25 -0800
 Subject: [PATCH] platform: add Raptor Blackbird support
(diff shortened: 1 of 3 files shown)


openpower/patches/blackbird-patches/skiboot/0001-platform-add-Raptor-Blackbird-support.patch
```diff
@@ -92,6 +90,15 @@ index 00000000..8817b8c5
 +	return true;
 +}
 +
++static void blackbird_exit(void)
++{
++	/* Let the IPL Observer know that we're done */
```
oohal: I thought tpearson was less-than-impressed by having this in skiboot? Could we do this using an init script in petitboot userspace?

stewartsmith:
On Mon, Feb 24, 2020, at 8:33 PM, oohal wrote:
> *@oohal* commented on this pull request.
> 
> In 
> openpower/patches/blackbird-patches/skiboot/0001-platform-add-Raptor-Blackbird-support.patch <https://github.com/open-power/op-build/pull/3471#discussion_r383653695>:
> 
> > @@ -92,6 +90,15 @@ index 00000000..8817b8c5
>  +	return true;
>  +}
>  +
> ++static void blackbird_exit(void)
> ++{
> ++	/* Let the IPL Observer know that we're done */
> I thought tpearson was less-than-impressed by having this in skiboot? 
> Could we do this using an init script in petitboot userspace?

He wants it to really signify that we're "done" IPLing and up at a petitboot UI. Although there's no way for petitboot to signal that to firmware currently, so we should really add a new OPAL call for it? AS is on FSP machines we kinda lie that IPL is done.

Without it, you get no fan control, which is certainly a non-trivial less than ideal share-a-home-office-with-your-partner harmony mode.

oohal:
> He wants it to really signify that we're "done" IPLing and up at a petitboot UI. Although there's no way for petitboot to signal that to firmware currently, so we should really add a new OPAL call for it?

Well, we've already got OPAL calls for reading and writing to the LPC space. The real problem is that they're only accessible to userspace via the debugfs interface at the moment. I suppose we could fix.

> AS is on FSP machines we kinda lie that IPL is done. Without it, you get no fan control, which is certainly a non-trivial less than ideal share-a-home-office-with-your-partner harmony mode.

Hmm, sounds like we need a driver that can signal to the outside world. Possibly via some kind of... operator panel.

sharkcz:
https://github.com/sharkcz/op-build/commit/44def594958d5614a2403ccc55ed94c519c38fe4 does the signaling from skiroot user-space, but it relies on debugfs. Personally I think reporting "IPL done" when exiting skiboot is acceptable until there is a better solution.

shenki:
We do this using an ipmi sensor on all the other BMC platforms. petitboot will signal that the OS has started as the last thing it does before kexec.

stewartsmith:
Superseded by https://github.com/open-power/op-build/pull/3554

PR #3476, opened 2020-02-26
buildroot: Support 5.5 kernels

shenki:
This adds support for 5.5 kernels to buildroot, but updates the
defconfigs in order to continue to use the 5.4 kernel for now.

### DIFF
--- a/openpower/configs/barreleye_defconfig
+++ b/openpower/configs/barreleye_defconfig
@@ -1,5 +1,6 @@
 BR2_powerpc64le=y
 BR2_powerpc_power8=y
+BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_5_4=y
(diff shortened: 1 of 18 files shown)


klauskiwi:
@shenki The immediate use I had for 5.5 (https://github.com/open-power/buildroot/pull/27) is no longer valid, so unless there's something else pressing us for a more modern kernel, I think we can wait for a buildroot release with 5.5 support..

Additionally, I think I'd like to keep using the latest kernel on upstream op-build whenever possible..

Thoughts?

shenki:
Buildroot are about to tag a release, and it does not come with support for 5.5. This would mean we're at least three months from being able to support 5.5, which would mean by then the current kernel is 5.6.

For that reason we have to patch in kernel support ourselves.

Additionally, we cannot move to a newer buildroot without fixing P8 hostboot (or removing the P8 platforms). I suggest we tag 2.5 and then make those changes.

Regarding the kernel that we use, yes, we can bump to 5.5 once this change has gone in.

PR #3482, opened 2020-02-27
skiboot v6.5-228-g82aed17a

opalci:
Oliver O'Halloran (3):
      core/platform: Add an explicit fast-reboot type
      core/fast-reset: Do not enable by default
      firenze-pci: Fix infinite loop in firenze_pci_add_loc_code()

Vasant Hegde (3):
      errorlog: Replace hardcode value with macro
      list: Add list_add_after()
      memory: Sort memory regions list


hegdevasant:
CI tests passed.. but for some reason it didn't update github PR.

Hence merging  this one.

-Vasant

PR #3554, opened 2020-04-09
Update Raptor Blackbird support

stewartsmith:
Notable changes:
- bump to latest Raptor 2.00 machine-xml
- Support IPL observer.

Skiboot patch has been sent upstream - https://patchwork.ozlabs.org/patch/1268510/
Hostboot rvpd patch has been sent upstream

and I'm doing a build+test of this final tree now (will be ~12-24 hours until I see it as I really should sleep rather than wait for op-build). But submitting this to have CI and feedback at least.


stewartsmith:
> and I'm doing a build+test of this final tree now (will be ~12-24 hours until I see it as I really should sleep rather than wait for op-build). But submitting this to have CI and feedback at least.

Update: it seems to work fine.

klauskiwi:
I think we now have pieces that have been merged in Skiboot, perhaps other places? Would you like to rebase this so we can take a look?

stewartsmith:
I'm having some fun as `host-localedef` now fails to build on Fedora 32, going to try and attack that this weekend.

stewartsmith:
Superseded by https://github.com/open-power/op-build/pull/3623

PR #3567, opened 2020-04-16
Bunch of minor fixes and cleanups for build-all-defconfigs.sh and update-defconfig.sh scripts

klauskiwi:


klauskiwi:
Will try again with a different patchset (#3570)

PR #3590, opened 2020-04-27
Update Explorer FW to v4.0

dcrowell77:
Updating to RC4 release

Signed-off-by: Dan Crowell <dcrowell@us.ibm.com>

PR #3594, opened 2020-04-28
Revert "Update Explorer FW to v4.0"

dcrowell77:
Reverts open-power/op-build#3590

Somehow my local build passed even though the release is missing the crucial binary...


klauskiwi:
> Somehow my local build passed even though the release is missing the crucial binary...

buildroot doesn't do a good job (or doesn't  do a job at all) at tracking dependencies in the case of rebuilds..

https://buildroot.org/downloads/manual/manual.html#full-rebuild

A good rule-of-thumb is to always `./op-build clean` if you are not rebuilding individual packages (and taking them out of the tree) or if you are changing makefiles..

BR2_CCACHE=y is your friend...

dcrowell77:
I would've thought that dirclean and rm -rf of the ocmb-explorer-fw tree would be enough... guess not.  I didn't have 4 hours to spend starting op-build from scratch.

PR #3623, opened 2020-05-13
Update Raptor Blackbird Support

stewartsmith:
Supersedes https://github.com/open-power/op-build/pull/3554

Notable changes:

- bump to latest Raptor 2.00 machine-xml
- Support IPL observer.

Skiboot patch has been sent upstream - https://patchwork.ozlabs.org/patch/1268510/
Hostboot rvpd patch has been sent upstream

Built and tested on my blackbird (but with modified buildroot so it builds on fedora 32 - see https://github.com/open-power/buildroot/pull/28 )


openpower/overlay/etc/init.d/S99signal-linux-complete
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+#
```
klauskiwi, line 2: @stewartsmith Is this init script supposed to be executed by every other platform as well?

stewartsmith:
It checks the device tree for the appropriate compat string, so it’s a no-op on other platforms, but will work anywhere the compat string is added.

klauskiwi:
Thanks for the clarification!

PR #3651, opened 2020-05-27
Fix Hostboot binutils usage

klauskiwi:
Hostboot needs to link against a static libfd.a and libiberty.a, but by
default buildroot will build the toolchain with shared libaries only, so
hostboot{-p8} hacks it's way through it by using the static libraries
from the binutils build directory, thus also depending on it's build.

In a scenario where we want to have a re-usable toolchain, any package
depending on host-binutils will force it to be rebuilt (even if we are
providing a perfectly fine binutils as part of our sdk).

Luckly, Hostboot use for those libraries is to build debugging and
tracing code using the HOST compiler (which is not a cross-compiler),
and since both Fedora and Ubuntu provides those packages, we can just
use them instead.


klauskiwi:
Thanks @dcrowell77 for noting this. The PR is still in RFC mode and I'm making adjustments here and there, but any thoughts are welcome.

If we agree with the concept, I'll propose the size increase for the BOOTKERNEL in the master openpower-pnor, and re-base this patch before merging.. 

Also, be advised that the hostboot-CI will fail because we don't have libbdl and libiberty installed there as far as I can tell..

klauskiwi:
@dcrowell77 @bofferdn @LF-Luis @ibmzach The CI build using containers (where we can specify the additional binutils-dev package neede) is passing, I think it's time we discuss this...

Your thoughts on this change?

ibmzach:
The only issue I can think of right now is that if we give up control of the libbfd version and a host upgrades its libbfd (which is an implementation library for GCC and as such is subject to breaking changes between compiler versions), then it could break tracehash or our linker in subtle ways. That chance may already be folded into the risk of upgrading compilers though.

klauskiwi:
> 
> 
> The only issue I can think of right now is that if we give up control of the libbfd version and a host upgrades its libbfd (which is an implementation library for GCC and as such is subject to breaking changes between compiler versions), then it could break tracehash or our linker in subtle ways. That chance may already be folded into the risk of upgrading compilers though.

Looks like libbfd is distributed specifically as a static library (and not a shared one) for the exact reason of it not having a stable API. So I guess the only defense against that is testing? Are we testing BFD/libiberty-dependant functions in our CI?

ibmzach:
In HB CI it's only being tested insofar as it is being used by tracehash and the linker. We don't have any testcases written specifically for it, and I'm not sure that it would help, because our CI is only run on systems that we control, whereas the bugs could/would manifest on other teams'/groups' machines with different compiler versions.

ibmzach:
Probably need @dcrowell77 or @bofferdn to comment before making CI changes

dcrowell77:
I can't think of any other options beyond just trying it I guess...

klauskiwi:
Got confirmation from Hai that these files were installed in the CI serves, so I rebased and force-pushed the patch. let's see how CI takes them.

klauskiwi:
retest this please

klauskiwi:
Just to give a quick status about this. Me and Zach worked this offline, and so far the conclusion has been that RedHat's Developer Toolset 8, that is used when building op-build on RHEL7 servers, has an incompatible backlevel version of binutils-devel package (2.30-55). We are requiring binutils-devel 2.31, otherwise Hostboot's `tracehash` application goes belly-up.

So right now we have no choice but to re-build binutils everytime, other than to carry it as part of a reusable toolchain

klauskiwi:
After a more recent round of discussions with @ibmzach and @dcrowell77, we concluded that the assumption originally used on this PR is not correct: Hostboot is extremely sensitive to the gcc + libbfd (binutils) combination, and not only for build-time, but also with potential consequences to runtime.

because of this, we cannot allow Hostboot to use the host's native compiler OR binutils. Instead, it needs to be pre-determined with the build configuration.

PR #3689, opened 2020-06-11
Update README

runlevel5:
Replace git URL with HTTPS URL because the latter does not require permission to clone


hegdevasant:
@ runlevel5  Patch looks good to me. Can you please add your s-o-b and send patch again?

-Vasant

klauskiwi:
Looks good, thanks

PR #3696, opened 2020-06-11
Update Linux Kernel to latest 4.4.227 version

klauskiwi:


klauskiwi:
The current CI build system is apparently not reliable to test this ancient branch, but I was able to build and sniff-test it locally, so merging this commit.

PR #3720, opened 2020-06-23
kernel: Move to Linux v5.4.48-openpower1

shenki:
This adds secure boot support backported from upstream and enables it
in the configuration.

Appearing in the backports is a patch to disable STRICT_KERNEL_RWX so
it drops out of the configuration.

Backported patches to support secureboot:

 powerpc/ima: Fix secure boot rules in ima arch policy
 powerpc/ima: Indicate kernel modules appended signatures are enforced
 powerpc/xmon: Allow listing and clearing breakpoints in read-only mode
 powerpc: Load firmware trusted keys/hashes into kernel keyring
 x86/efi: move common keyring handler functions to new file
 powerpc: expose secure variables to userspace via sysfs
 powerpc/powernv: Add OPAL API interface to access secure variable
 powerpc/ima: Update ima arch policy to check for blacklist
 ima: Check against blacklisted hashes for files with modsig
 certs: Add wrapper function to check blacklisted binary hash
 ima: Make process_buffer_measurement() generic
 powerpc/ima: Define trusted boot policy
 powerpc: Detect the trusted boot state of the system
 powerpc/ima: Add support to initialize ima policy rules
 powerpc: Detect the secure boot mode of the system

PowerPC related fixes:

 powerpc/64s: Save FSCR to init_task.thread.fscr after feature init
 powerpc/64s: Don't let DT CPU features set FSCR_DSCR
 powerpc/kasan: Fix shadow pages allocation failure
 powerpc/kasan: Fix issues by lowering KASAN_SHADOW_END
 powerpc/fadump: Account for memory_limit while reserving memory
 powerpc/fadump: consider reserved ranges while reserving memory
 powerpc/fadump: use static allocation for reserved memory ranges
 powerpc/mm: Fix conditions to perform MMU specific management by blocks on PPC32.
 powerpc/spufs: fix copy_to_user while atomic
 sched/core: Fix illegal RCU from offline CPUs
 powerpc/ptdump: Properly handle non standard page size
 powerpc/xive: Clear the page tables for the ESB IO mapping
 bpf: Support llvm-objcopy for vmlinux BTF
 powerpc/xmon: Restrict when kernel is locked down
 powerpc/powernv: Avoid re-registration of imc debugfs directory
 powerpc/64s: Disable STRICT_KERNEL_RWX
 powerpc: Remove STRICT_KERNEL_RWX incompatibility with RELOCATABLE
 powerpc/mm: Fix CONFIG_PPC_KUAP_DEBUG on PPC32
 powerpc/kuap: PPC_KUAP_DEBUG should depend on PPC_KUAP
 powerpc/setup_64: Set cache-line-size based on cache-block-size
 Revert "powerpc/64: irq_work avoid interrupt when called with hardware irqs enabled"


openpower/configs/linux/skiroot_defconfig
```diff
@@ -209,7 +209,6 @@ CONFIG_IPMI_DEVICE_INTERFACE=y
 CONFIG_IPMI_POWERNV=y
 CONFIG_IPMI_WATCHDOG=y
 CONFIG_HW_RANDOM=y
-CONFIG_TCG_TPM=y
```
klauskiwi, line 212: Is this intentional? My understanding is that CONFIG_TCG_TIS_I2C_NUVOTON depends on CONFIG_TCG_TPM, so the next value will be masked if we make olddefconfig?

shenki: IMA selects TCG_TPM:
```
security/integrity/ima/Kconfig:12:      select TCG_TPM if HAS_IOMEM && !UML
```

klauskiwi: Thanks for the clarification!

PR #3814, opened 2020-08-29
Fixes package list for Ubuntu 20.04 CI

### LINKED ISSUE #3808: [CI] Ubuntu 24.04 docket setup fails
Commit `34b9a2dc7` added support for ubuntu20.4. But its failing for me with below error.

17:14:57  ---> Running in 23853d3465df
17:14:58 Reading package lists...
17:14:58 Building dependency tree...
17:14:58 Reading state information...
17:14:58 Package python-matplotlib is not available, but is referred to by another package.
17:14:58 This may mean that the package is missing, has been obsoleted, or
17:14:58 is only available from another source
17:14:58 
17:14:58 [91mE: Package 'python-matplotlib' has no installation candidate

-Vasant

klauskiwi:
Commit 34b9a2dc7 moved the CI to Ubuntu 20.04 Focal but was
still depending on python-matplotlib and python-numpy for it's
'release' steps (where make legal-info and graphs are built),
while those are no longer available in Focal.

Replace them for python3-matplotlib and python3-scipy, also
adjusting python to python3-is-python so that things work
again.

Signed-off-by: Klaus Heinrich Kiwi <klaus@linux.bnet.ibm.com>
----------
I missed this since the Ci was skipping silently for Ubuntu 20.04 containers. That should be fixed as well.

Fixes #3808

### DIFF
--- a/ci/Dockerfile/ubuntu2004
+++ b/ci/Dockerfile/ubuntu2004
@@ -2,9 +2,9 @@ FROM ubuntu:20.04
 # Don't ask. Ubuntu is awful.
 RUN sed -e 's/main$/main universe/' --in-place=orig /etc/apt/sources.list
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -yy cscope ctags libz-dev libexpat-dev \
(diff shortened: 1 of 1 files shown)


klauskiwi:
retest this please

PR #3906, opened 2020-10-27
Move mowgli-xml to github.com/open-power

klauskiwi:
Move mowgli-xml repo from ibm-op-release to open-power github organization


klauskiwi:
@LuluTHSu @NicholeWang FYI

PR #3912, opened 2020-11-01
README: Update "Building an image" example

gkurz:
POWER8 targets were removed by commit 40f804058414. Use a POWER9
one for the example.


klauskiwi:
Thanks!

PR #3927, opened 2020-11-10
Buildroot bump to buildroot-2020.08.2

klauskiwi:


klauskiwi:
@shenki Joel, can you take a look and see if it's good enough fort us to push it into github.com/open-power/buildroot as a 2020.08.1-op-build branch?

The CI issues on RHEL7 are false-positives and are being worked with the CI team already. And because of that it didn't trigger the IPL test, but I tested for sanity on a Witherspoon DD2.3 and it looks alright..

klauskiwi:
Ps: if this looks good, I'd push the Pnor changes upstream, and then integrate all that in a op-build PR

klauskiwi:
created a clean #3941 to supersede this one

Issue #3945, opened 2020-11-29
Doesn't build on ppc64el


stewartsmith:
On v2.6 tag and on master (v2.6-35-g45db0624):
```
$ op-build blackbird_defconfig
$ op-build
.....
>>> host-libopenssl 1.1.1d Extracting
gzip -d -c /home/stewart/op-build/dl/libopenssl/openssl-1.1.1d.tar.gz | /home/stewart/op-build/output/host/bin/tar --strip-components=1 -C /home/stewart/op-build/output/build/host-libopenssl-1.1.1d   -xf -
>>> host-libopenssl 1.1.1d Patching

Applying 0001-Dont-waste-time-building-manpages-if-we-re-not-going.patch using patch: 
patching file Configurations/unix-Makefile.tmpl
Hunk #1 succeeded at 491 (offset 223 lines).

Applying 0002-Reproducible-build-do-not-leak-compiler-path.patch using patch: 
patching file crypto/build.info

Applying 0003-Introduce-the-OPENSSL_NO_MADVISE-to-disable-call-to-.patch using patch: 
patching file crypto/mem_sec.c
>>> host-libopenssl 1.1.1d Configuring
[...]

PR #3946, opened 2020-11-29
Add perl dependencies for building on Fedora 33

stewartsmith:
Add perl FindBin, English, Time dependencies for Fedora
    
Otherwise host-libopenssl fails to build, saying ppc64el isn't
supported, but a couple of lines up, hints at the perl module.
    
Hostboot also needs the English and Time::localtime modules.

This should also be cherry-picked to v2.6 at least.


klauskiwi:
Good catch, thanks. I'll pull it into the 2.6 release as well.

Fixes #3945

Issue #3003, opened 2019-07-16
pseries machine


shenki:
We have the Linux kernel `pseries_skiroot_defconfig and the buildroot `pseries_defconfig` in the tree. These are occasionally used by kernel developers (@mpe) to build a petitboot for pseries.

The config hasn't seen much love, with @sammj the only one to have ever touched the kernel config.

There is a idea to upstream our petitboot configuration to buildroot. When we do this, people that want to do a buildroot+petitboot+kernel can use that instead. When this work happens, we can consider removing the version in op-build.

shenki:
The correct fix is to drop the pseries defconfig. It is no longer used.

Issue #1389, opened 2017-08-31
in petitboot fsck does not work as expected


bssrikanth:
In petitboot I see fsck command present but that does not work as expected..
>fsck
fsck (busybox 1.26.2, 2017-08-24 11:20:16 PDT)
fsck: fsck.ext2: No such file or directory

Samuel Mendoza-Jonas [5:35 AM] 
hmm, no i expect we should at least have .ext2 support
ah looks like busybox doesn't package it by default, i'll have a look

So this issue. Thank you.

Below are the levels of host fw, petitboot etc.:

 Product Name          : OpenPOWER Firmware
 Product Version       : open-power-SUPERMICRO-P9DSU-V0.12-20170824
 Product Extra         : 	op-build-da78c29-dirty
 Product Extra         : 	skiboot-f5d9049-pa7d7d42
 Product Extra         : 	hostboot-75333b1
 Product Extra         : 	linux-4.12.5-openpower1-pbf81229
 Product Extra         : 	petitboot-v1.5.1-p11bf40f
 Product Extra         : 	machine-xml-a134897
 Product Extra         : 	occ-71b5f68
 Product Extra         : 	sbe-4a62126

bssrikanth:
@sammj please take a look. Thank you.

stewart-ibm:
Srikanth Aithal <notifications@github.com> writes:
> In petitboot I see fsck command present but that does not work as expected..
> # fsck
> fsck (busybox 1.26.2, 2017-08-24 11:20:16 PDT)
> fsck: fsck.ext2: No such file or directory
>
> Samuel Mendoza-Jonas [5:35 AM] 
> hmm, no i expect we should at least have .ext2 support
> ah looks like busybox doesn't package it by default, i'll have a look

I think we may have had it way back in the pre-op-build Tuleta days as
we have more room to play with there (16mb for initramfs).

My guess is the size increase is the big problem.

-- 
Stewart Smith
OPAL Architect, IBM.

sammj:
I did a quick build comparison, looks like BR2_PACKAGE_UTIL_LINUX_FSCK adds 108KB. So it fits, but it depends on whether we're going to need that space for something else.

sammj:
Actually that could be up to 500KB depending on the combination of util-linux and e2fsprogs options we use.

stewart-ibm:
Sam Mendoza-Jonas <notifications@github.com> writes:
> Actually that could be up to 500KB depending on the combination of
[...]

PR #3964, opened 2020-12-11
[RFC] Support for builtin kernels

klauskiwi:
There are three main motivations:
 * consolidate around one single skiroot_defconfig (i.e., https://github.com/open-power/op-build/issues/3003)
 * Allow for more complete skiroot (i.e., https://github.com/open-power/op-build/issues/1389)
 * Still keeping each of PAYLOAD / Skiboot.lid and BOOTKERNEL / rootfs.lid below 16MiB for future compatibility with FSP systems

This still doesn't impede us from creating a secondary partition / rootfs eventually either as an overlay or another mountpoint for Skiroot, but this is a good first step I think.

@stewartsmith @shenki thoughts?


stewartsmith:
This will *probably* slow down boot as Hostboot pretty much stops doing anything except bringing in `PAYLOAD`, and this will dramatically increase that time. It also makes it hard to replace the kernel independently of replacing skiboot.

klauskiwi:
> 
> 
> This will _probably_ slow down boot as Hostboot pretty much stops doing anything except bringing in `PAYLOAD`, and this will dramatically increase that time. It also makes it hard to replace the kernel independently of replacing skiboot.

This shouldn't prevent defconfigs from breaking up between PAYLOAD, BOOTKERNEL and even ROOTFS if they want to, but perhaps we could consolidate on a single skiroot_defconfig that works for all the platforms (including fsp)?

  -Klaus

hegdevasant:
Couple of points:
- As @stewartsmith mentioned this will increase the boot time.
- In system platform we have hard limit on individual firmware image (aka LID size). So we will hit all sort of size limit issue for no good reason. 
- Replacing individual image (LID) for testing becomes complicated. 
 
Finally even if we want to have single file for particular platform then we should restrict changes  to that platform. 

-Vasant

klauskiwi:
>     * As @stewartsmith mentioned this will increase the boot time.

I did a few reboot tests on a witherspoon system. Petitboot-to-petitboot reboot is taking around 146s on with the current layout, and 145-150s on the new layout. So I think it's within acceptable range.
 
>     * In system platform we have hard limit on individual firmware image (aka LID size). So we will hit all sort of size limit issue for no good reason.

I don't understand this comment at all - could you clarify? The objective is exactly to allow more Skiroot content on platforms that are restricted in LID size (by allowing us to use a 16MB PAYLOAD LID plus 16MB BOOTKERNEL and/or ROOTFS LID. So what do you mean  by _no good reason_?

>     * Replacing individual image (LID) for testing becomes complicated.

Agreed. A Skiboot developer would need to have a skiroot kernel around to embed it every time it needs to be replaced on the target system.  I'm sure we can use one of the several CI jobs running daily on op-build to grab a kernel if necessary.

> Finally even if we want to have single file for particular platform then we should restrict changes to that platform.
> -Vasant

I'd rather use this to allow us a single kernel skiroot_defconfig (see #3003) and hopefully simplify maintenance of it, but I can agree to break it down into fixing the broken SKIBOOT_EMBED_PAYLOAD config first, and discussing the changes to platforms defconfigs later.

shenki:
This behavior is a regression; built in kernels was only used during skiboot bringup before we had the ability to talk to the flash. I suggest we drop support for builtin kernels.

hegdevasant:
> 
> I did a few reboot tests on a witherspoon system. Petitboot-to-petitboot reboot is taking around 146s on with the current layout, and 145-150s on the new layout. So I think it's within acceptable range.

Depends. Our goal is to boot system as fast as possible. This defeats our goal. 

> 
> I don't understand this comment at all - could you clarify? The objective is exactly to allow more Skiroot content on platforms that are restricted in LID size (by allowing us to use a 16MB PAYLOAD LID plus 16MB BOOTKERNEL and/or ROOTFS LID. So what do you mean by _no good reason_?

Today skiboot is very small (less than 3MB). If any reason we start increasing SKIBOOT size then it complicates ..because kernel would have used remaining space. 
I thought 16MB BOOTKERNEL + 16MB ROOTFS is sufficient for skiroot. Are we crossing that limit? 

> 
> > ```
> > * Replacing individual image (LID) for testing becomes complicated.
> > ```
> 
> Agreed. A Skiboot developer would need to have a skiroot kernel around to embed it every time it needs to be replaced on the target system. I'm sure we can use one of the several CI jobs running daily on op-build to grab a kernel if necessary.

So grab kernel , rebuild skiboot with kernel and flash it? I don't think this is right thing to do.

-Vasant

klauskiwi:
> 
> 
> > I did a few reboot tests on a witherspoon system. Petitboot-to-petitboot reboot is taking around 146s on with the current layout, and 145-150s on the new layout. So I think it's within acceptable range.
> 
> Depends. Our goal is to boot system as fast as possible. This defeats our goal.

Show me, because my evidence so far hasn't shown me otherwise.

> > I don't understand this comment at all - could you clarify? The objective is exactly to allow more Skiroot content on platforms that are restricted in LID size (by allowing us to use a 16MB PAYLOAD LID plus 16MB BOOTKERNEL and/or ROOTFS LID. So what do you mean by _no good reason_?
> 
> Today skiboot is very small (less than 3MB). If any reason we start increasing SKIBOOT size then it complicates ..because kernel would have used remaining space.
> I thought 16MB BOOTKERNEL + 16MB ROOTFS is sufficient for skiroot. Are we crossing that limit?
> 
yes. There are a number of platforms already carrying modified pnor layouts with larger BOOTKERNEL (specially since we're not using ROOTFS on pnor layouts). 

> > > ```
> > > * Replacing individual image (LID) for testing becomes complicated.
> > > ```
> > 
> > 
> > Agreed. A Skiboot developer would need to have a skiroot kernel around to embed it every time it needs to be replaced on the target system. I'm sure we can use one of the several CI jobs running daily on op-build to grab a kernel if necessary.
> 
> So grab kernel , rebuild skiboot with kernel and flash it? I don't think this is right thing to do.

That depends on your point-of-view, after all is just a question of who is embedding the kernel and where.. (i.e., someone has to do it)

> -Vasant

Considering the feedback here, my suggestion then is to remove the pseries_skiroot_defconfig (#3990) and put the default skiroot_defconfig on the best diet possible (ahem @shenki - still waiting on https://lists.ozlabs.org/pipermail/openpower-firmware/2020-December/000600.html), while still keeping the support for an embedded skiroot kernel on skiboot for the time being. So we can close this now.

PR #4028, opened 2021-03-03
ci: Add Fedora33 to default CONTAINERS being built

klauskiwi:
Commit 23a11fb "ci: Move Fedora32 Dockerfile to Fedora33" moved the CI
build script to use Fedora33, but this was actually not the default in
the main build.sh script when no Distros were being specified.


klauskiwi:
retest this please

PR #4231, opened 2021-07-27
[RFC] Introduce BR2_SKIBOOT_EXTRA_CONFIG_OPTIONS

hegdevasant:
P9 OpenPower systems has defined PAYLOAD (skiboot LID) size as 512K (see [1]).
This worked fine so far as our signed LID never cross this limit. Recently
in some cases signed skiboot LID is crossing 512K limit. Hence this patches
add new skiboot config option to pass extra config (like build without FSP
code) in system config.

Also this patch adds below config to OpenPower P9 system configs. This
will allow us to compile skiboot without FSP code which helps to reduce
skiboot LID size ~50K.
  BR2_SKIBOOT_EXTRA_CONFIG_OPTIONS="CONFIG_FSP=0"

[1] https://github.com/open-power/pnor/blob/master/p9Layouts/defaultPnorLayout_64.xml


hegdevasant:
retest please

shenki:
Further investigation suggests the layout is 1MB for all the supported systems these days. This suggests we don't need the different skiboot configuration.

Could we avoid it for now to keep the testing matrix smaller?

hegdevasant:
We had a internal discussion on this PR and we concluded that:
  - For P10 - we will increase PAYLOAD size to 1MB (I have posted fix to pnor repo)
  - For P9 - We have already reverted and latest pnor repo has size of 1MB
  - For P8 - op-build master doesn't support P8 platforms. Hence we are good. If someone manually tries to flash upstream skiboot on P8 system then it will break. We have to use CONFIG_FSP=0 option. We think its ok as its not officially supported and hardly few people uses/tests upstream skiboot on P8 system.

-Vasant

PR #4459, opened 2021-12-07
Move Rainier/Everest XML Pointers to GHE

IlyaSmirnov91:
Signed-off-by: Ilya Smirnov <ismirno@us.ibm.com>


lili-lilili:
Move these xml files to a private repositories? why?

PR #4513, opened 2022-01-20
Allow "Security Version" To Be Set For A Given Configuration

mabaiocchi:
This commit allows for a given configuration/system to set an optional
"Security Version" value.  It specifically sets the value of the p10ebmc_defconfig
configuration to zero.  This value then gets passed into the openpower-pnor-p10
package and is built into the partitions/lids via the signing utilities.  This
commit points to the openpower PNOR repo's master-p10 commit that has the
needed support for this.

Signed-off-by: Mike Baiocchi <mbaiocch@us.ibm.com>


IlyaSmirnov91 (approved):

mabaiocchi:
FYI I retired the failing "IBM HW Tests".  It looks like it restarted correctly.

PR #4556, opened 2022-02-23
DO NOT MERGE Op build update 1021 2 23 2022 fix for hcode

bofferdn:


bofferdn:
This is purely proving that Jason Albert can make an equivalent change in ekb and fix the ubuntu HCODE compile

bofferdn:
retest this please

PR #4573, opened 2022-03-03
bump recipe to pull in ssh-agent support

Joshua-Andersen1:
Signed-off-by: Joshua Andersen <joshua.andersen1@ibm.com>


fspbld:
@bofferdn  --- this is per the email that Josh sent last week. Basically, the issue is that the version of the sign framework is so old that ssh-agent is not supported.   So, when it comes time to SB sign an op-build tree, it prompts for a passphrase every time. It was tolerable in p9 because it was only 16 prompts. In p10 it is more than 3x that... (one set for each model + the extra calls sbe has to do).

IlyaSmirnov91:
Nick approved - merging.

PR #4615, opened 2022-03-24
skiboot: Use open-opwer/skiboot.git again

shenki:
By defaulting to BR2_SKIBOOT_CUSTOM_GIT, the custom URL overrode the
default URL for the skiboot repo, forcing the need to have a defualt
custom URL too.

This which was used during internal p10 bring up to force the URL to an
internal repository. There's no need to do it now, and github.com is
moving away from the git:// protocol, so remove the custom default so we
get buildroot's normal URL construction from SKIBOOT_SITE in skiboot.mk.

Overriding the URL can still be done via the defconfig where it is
useful.


IlyaSmirnov91:
ok to test

PR #4617, opened 2022-03-24
buildroot: Update to 2021.02.10 stable point release

shenki:
This brings in the latest point release for the buildroot branch we're
on, which is mostly securty fixes to the packages and tools.

Importantly it includes a fix for localedef when the host has GCC 11,
such as RHEL9.


shenki:
This is the same as https://github.com/open-power/op-build/pull/4616, but for the master-p10 branch.

IlyaSmirnov91:
ok to test

shenki:
33546240 = 16773120 * 2. I wonder what's going on there?

On p10, we have no real pnor size constraints. Can we simply make the bootkernel partition larger?

IlyaSmirnov91:
> 33546240 = 16773120 * 2. I wonder what's going on there?
> 
> On p10, we have no real pnor size constraints. Can we simply make the bootkernel partition larger?

Our PNOR script doubles the size if it goes over the limit even a little bit, so the real size isn't necessary twice as large. 

Yes, we can increase the size of BOOTKERNEL in PNOR to accommodate these changes.

bofferdn:
Well, the BOOTKERNEL has to fit into 16 MB of space to be able to fit into a single LID, which has 16 MB as the architected  upper limit for in-band firmware update (and there might be some overhead that takes away from even that).   I don't think we have the liberty of getting much bigger, it's already very close to the 16 MB edge, and I don't know the exact number required to accommodate the LID overhead.

bofferdn (changes requested):
see main comment re: the LID size; we'll need to work through that size restriction.

dcrowell77:
So who has the ball here on shrinking BOOTKERNEL down (since it seems like this is a required update at some point)?  @shenki

bofferdn:
> So who has the ball here on shrinking BOOTKERNEL down (since it seems like this is a required update at some point)? @shenki

Historically that has been up to the component that creates the BOOTKERNEL content.  Either enough 'fat' will need to be trimmed out of there to stay below the 16 MB max size, or the content will need to be split into multiple LIDs and loaded sequentially.

runlevel5:
@IlyaSmirnov91 I think we could close this PR now

IlyaSmirnov91:
> @IlyaSmirnov91 I think we could close this PR now

Is this PR not relevant anymore? This one is for master-p10 branch instead of master.

runlevel5:
I do not think it is relevant anymore. At least we should rebase p10 with current master branch

IlyaSmirnov91:
Closing.

PR #4620, opened 2022-03-24
Kernel 5.10.105

shenki:
Bump the kernel to the latest stable, and bring with it an update so networking works under petitboot again.


IlyaSmirnov91:
ok to test

PR #4624, opened 2022-03-24
4k blank WOF data file

chenduibm:
Set a worse case scenario of creating a 4k blank WOF data file for when the WOF input is not supplied

Signed-off-by: chenduibm <duchen@us.ibm.com>


bofferdn (approved):
The indent issue seems not real when I look at the raw files.

bofferdn:
retest this please

PR #4628, opened 2022-03-25
p10ebmc: Update everest-xml repository

### LINKED ISSUE #4627: p10ebmc_defconfig fails to build due to everest-xml
```
$ git clone git@github.com:open-power/everest-xml.git
Cloning into 'everest-xml'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

If a build machine does not have a ssh key set up to log in to github, the build will fail (even though this repo is public).

We should be using https:// to access github, per https://github.blog/2021-09-01-improving-git-protocol-security-github/

shenki:
Remove the CUSTOM_GIT variables so we fall back to the default github
path. This in turn uses the _SITE variable to construct the repository
path, so we get https://github.com URLs.

Fixes #4627.


IlyaSmirnov91:
ok to test

PR #4650, opened 2022-04-01
Patch for DIMM number expansion

dcrowell77:
Bumping Hostboot up to include a commit that expands the DIMM
numbering.  This requires a coreq HWP change that is not yet
merged so pulling that in as a patch for now.


dcrowell77:
Note to admins - Check with build team before merging this as there are other critical fixes in flight that we want in front of this one in case of issues.

PR #4808, opened 2022-06-22
Set The Secure Version to 2 for P10 eBMC Builds

mabaiocchi:
This commit sets the variable BR2_OPENPOWER_SIGNED_SECURITY_VERSION
in the p10ebmc_defconfig configuration to "2" from "0".  This new
value of 2 will be built into the secure headers of the lids.

Once a driver with this Secure Version of 2 is installed on a system,
a user will not be able to boot a driver with a Secure Version less
than 2 with security enabled.

Signed-off-by: Mike Baiocchi <baiocchi@ibm.com>


mabaiocchi:
@fspbld can you please go to the "File Changes" tab, and then click on the "Review Changes" green button on the right and select "Request changes" ? 

Ideally, that should more clearly show a "-1" review on this, but still let CI continue.

mabaiocchi:
@bofferdn @cjengel Please review this critically important one line of code :)

PR #4868, opened 2022-07-19
Add occStringFile to PNOR scratch dir

cjcain:
Adding occStringFile so that a lid file can be created.
Lid: 81e00687

CQ: SW554564
Signed-off-by: Chris Cain <cjcain@us.ibm.com>


fspbld:
ok to test

PR #4895, opened 2022-08-04
SBE-P10: Update the call python to python2 for Red Hat 8 migration

velozr:
For Red Hat 8, there will be no default /bin/python only /bin/python2
and /bin/python3.  For Red Hat 7 /bin/python is just a symlink to python2,
therefore all calls to python need to be changed to python2 which is
safe to do.

Signed-off-by: Roland Veloz <rveloz@us.ibm.com>


fspbld:
ok to test

bofferdn (approved):

PR #4939, opened 2022-09-07
Remove deprecated everest.config file

dcrowell77:
All P10 eBMC systems should use the common p10ebmc.config file.

Signed-off-by: Daniel Crowell <dcrowell@us.ibm.com>


dcrowell77:
ok to test

PR #5081, opened 2022-11-29
Update rainier-xml Pointers

IlyaSmirnov91:
There is an issue with automatic PR creator because the pointers to rainier-xml repo are incorrect. This commit updates the pointers to rainier-xml to the latest commit in the repo to fix that issue.

Signed-off-by: Ilya Smirnov <ismirno@us.ibm.com>


IlyaSmirnov91:
The same as #5080

PR #5147, opened 2023-01-18
OP-BUILD Failure Hardening

debmc:
Need to chain the multiple shell commands together with && for create_pnor_image.pl so that any failures in called scripts (e.g. die, etc) are properly caught and surfaced.

Without chaining the shell commands together with && the success of the operation to create_pnor_image.pl may fail silently.

Exposure to this type of problem was detected with the newly introduced buildSPDImages.pl script which was stress tested by overflowing the PSPD image which is built outside the space constraints of the PNOR logic (checkSpaceConstraints).

Signed-off-by: Deb McLemore <debmc@us.ibm.com>


fspbld:
ok to test

debmc:
retest this please

PR #5282, opened 2023-03-27
p10 Hcode: Partial support for PM eLog trace parsing.

premsjha:
Commit copies trexString file from build directory to a PNOR staging directory as a lid.


premsjha:
commit not signed off

PR #5284, opened 2023-03-27
p10_hcode: Partial support for parsing of traces in PM elog

premsjha:
Commit copies PM trexStringFiles from build directory to PNOR stage directory as LID.

### DIFF
--- a/openpower/package/openpower-pnor-p10/openpower-pnor-p10.mk
+++ b/openpower/package/openpower-pnor-p10/openpower-pnor-p10.mk
@@ -3,7 +3,6 @@
 # openpower_pnor_p10
 #
 ################################################################################
-
 OPENPOWER_PNOR_P10_VERSION ?= 38b8d4759092a42d77a4f28939a0e3730f00e1f3
 OPENPOWER_PNOR_P10_SITE ?= $(call github,open-power,pnor,$(OPENPOWER_PNOR_P10_VERSION))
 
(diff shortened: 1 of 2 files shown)


fspbld:
ok to test

dcrowell77:
Looks like there is a glitch in the CI logic.  The "Ubuntu ppc64le" job got assigned to one of the x86 ubuntu machines again.

openpower/package/openpower-pnor-p10/openpower-pnor-p10.mk
```diff
@@ -3,7 +3,7 @@
 # openpower_pnor_p10
 #
 ################################################################################
-
+HCODE_STAGE_DIR = $(STAGING_DIR)/../../../build/hcode-p10-*opmst10/output/images
```
dcrowell77, line 6: The typical way of handling this sort of thing is for each package to deliver content into the main staging area, rather than having the op-build code contain package-specific paths.  That way the high-level op-build layer remains agnostic.  If you look at openpower/package/hcode-p10/hcode-p10.mk it already delivers the images into the $(STAGING_DIR)/hcode/ directory.  You should modify that file instead of this one.  Even better would be to create a layer of indirection where  hcode-p10.mk calls some kind of generic script inside the hcode src tree that copies all of the useful information out to the staging area.

premsjha: Okay

dcrowell77:
retest this please

PR #5290, opened 2023-03-29
Hostboot Build Planar SPD cleanup

debmc:
Update some variable names to more clearly identify the usage.

PFHB-5


fspbld:
ok to test

PR #5351, opened 2023-04-25
phal: pdata: Bumped 5c0c448d43..e28301d3e6f

devenrao:
Marri Devender Rao (1):
      dt-api : do not log error trace if attribute not found in the target

Change-Id: I9b57d80e175e217898c4cd100621168874ee44f5


fspbld:
ok to test

PR #5433, opened 2023-05-25
p10_hcode: Partial support for parsing of traces in PM elog

premsjha:
Commit copies PM trexStringFiles from build directory to PNOR stage directory as LID.


dcrowell77:
@fspbld - FYI, this is a port of the changes that are in master-p10.

PR #5443, opened 2023-05-29
phal: pdata: Bumped e28301d3..298b7970

deepakala-k:
Joel Stanley (2):
      configure: Run autoupdate
      configure: Make aarch64 equivalent to arm

deepakala-k (1):
      scripts: Use numerical sorting if applicable


devenrao:
looks good to me

fspbld:
ok to test

PR #5599, opened 2023-08-04
Add IBM Dockerfile

danlarson-ibm:


ci/ibm/podman_local.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# allows user to volume mount a op-build repo
+opbuild_dir=${1:-${HOME}/op-build}
+# uses git branch name by default 
+tag_name=${2:-op-build:$(git rev-parse --abbrev-ref HEAD)}
+
+working_dir=/home/$USER/op-build
+#exit
+#
+podman build --build-arg UID=$UID --build-arg GID=$(id -g) --build-arg USER=$USER -t $tag_name -f ci/ibm/Dockerfile ci/ibm
```
runlevel5, line 12: What do you think about renaming `ibm/Dockerfile` to `ibm/Containerfile`? IMHO the `Containerfile` is a better choice just because we are not using `docker` at all here

danlarson-ibm:
start jenkins

PR #5679, opened 2023-09-12
Sync git submodule

runlevel5:
## Changes

* git submodule sync the buildroot (my bad I did not do so in the [PR](https://github.com/open-power/op-build/pull/5629)
* Opt in the new default GCC 11
* Bump loadkeys to 2.5.1 to avoid compilation error with GCC 11

PR #5629, opened 2023-08-21
buildroot: Bump to version 2023.02-op-build

runlevel5:
## Context

There have been few attempts to bump to newer linux 5.10.x and buildroot. Unfortunately most of them have failed builds on CI. So I'd like to do another attempt but starting with bumping to newer buildroot first.

## Changes

* Bump to buildroot version 2023.02-op-build

## NOTES

I can successfully compile `blackbird` firmware on Ubuntu 22.04 ppc64le

PR #5685, opened 2023-09-15
Revert "buildroot: Bump to version 2023.02-op-build"

runlevel5:
## What

Reverts open-power/op-build#5629

## Why

It is totally my mistake for not running `git submodule update` which means the `buildroot` is still pointing to the `2021.02-op-build`.

I would like to revert that changes and will work on a [new PR](https://github.com/open-power/op-build/pull/5679) to bump buildroot to `2023.02`.

cc @IlyaSmirnov91

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "buildroot"]
 	path = buildroot
-	branch = 2023.02-op-build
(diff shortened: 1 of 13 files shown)


fspbld:
ok to test

PR #5729, opened 2023-10-11
Copy Odyssey debug tools

sumitk56:
Change-Id: Ia9ebee4aba179fa3e8ad5ebeea11c2c549dd7ecd

### DIFF
--- a/openpower/package/sbe-odyssey/sbe-odyssey.mk
+++ b/openpower/package/sbe-odyssey/sbe-odyssey.mk
@@ -29,8 +29,10 @@ endef
 
 define SBE_ODYSSEY_INSTALL_IMAGES_CMDS
 	mkdir -p $(STAGING_DIR)/ody_binaries
+	mkdir -p $(STAGING_DIR)/poz_debug_tools
         cp $(STAGING_DIR)/ody-pak-files/gen/final/boot.pak $(STAGING_DIR)/ody_binaries/
         cp $(STAGING_DIR)/ody-pak-files/gen/final/rt.pak $(STAGING_DIR)/ody_binaries/
+        cp $(@D)/images/odyssey/odyssey_sbe_debug_DD1.tar.gz $(STAGING_DIR)/poz_debug_tools/
 endef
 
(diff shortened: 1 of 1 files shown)


sumitk56:
ok to test

fspbld:
ok to test

openpower/package/sbe-odyssey/sbe-odyssey.mk
```diff
@@ -29,8 +29,10 @@ endef
 
 define SBE_ODYSSEY_INSTALL_IMAGES_CMDS
 	mkdir -p $(STAGING_DIR)/ody_binaries
+	mkdir -p $(STAGING_DIR)/ody_debug_tools
```
RameshIyyar, line 32: I would suggest using "poz" instead of "ody", since all tools we are planning to make it common and some simics scripts already made it common.

What about the dd level?

openpower/package/sbe-odyssey/sbe-odyssey.mk
```diff
@@ -29,8 +29,10 @@ endef
...
         cp $(STAGING_DIR)/ody-pak-files/gen/final/boot.pak $(STAGING_DIR)/ody_binaries/
         cp $(STAGING_DIR)/ody-pak-files/gen/final/rt.pak $(STAGING_DIR)/ody_binaries/
+        cp $(@D)/images/odyssey/odyssey_sbe_debug_DD1.tar.gz $(STAGING_DIR)/ody_debug_tools/
```
RameshIyyar, line 35: Are you planning to push golden string files in the separate PR?

openpower/package/sbe-odyssey/sbe-odyssey.mk
sumitk56: yeah will make poz. dd level will see later

openpower/package/sbe-odyssey/sbe-odyssey.mk
sumitk56: yeah will checkout

PR #5828, opened 2023-12-20
phal: pdata: Bumped 85e3c44..722e521

deepakala-k:
deepakala-k (2):
      Add odyperv targets to device tree
      Fix index duplication for memports and DIMMs


fspbld:
ok to test

PR #5867, opened 2024-01-24
Adjust Odyssey SBE Build Command

dcrowell77:
Using the 'sbe workon' command spawns a new shell which can have some bad side effects to some of the automated build scripts. This change switches over to using an inline script for the build instead.


dcrowell77:
All signs point to this working so going to merge.

PR #5909, opened 2024-02-19
Add BR2_OPENPOWER_SIGNED_SECURITY_VERSION into sbe-p10 and sbe-odyssey makefiles

mabaiocchi:
This commit converts the buildroot variable BR2_OPENPOWER_SIGNED_SECURITY_VERSION
    into a local environment variable in specific spots of the sbe-p10 and
    sbe-odyssey makefiles.  This ensures that those packages are built with the
    same value as the rest of the build.


mabaiocchi:
Closing this as I didn't mean to pick up the other 3 commits

PR #5920, opened 2024-02-23
Remove custom git references for bonnell-xml

dcrowell77:
Removed one more leftover reference to openbmc that should hopefully let me remove the custom git indicators.


mabaiocchi (approved):
bahh - I should have caught this the first time

PR #5956, opened 2024-03-13
petitboot: Update to v1.15

rarbab:
Update to the latest petitboot for bug fixes. Most notably, it fixes a bootloader config parsing issue causing the discovery process to crash when it encounters installations of recent Fedora CoreOS.


fspbld:
ok to test

PR #6123, opened 2024-05-24
Add BR2_OPENPOWER_SIGNED_SECURITY_VERSION into sbe-p10 makefile

mabaiocchi:
This commit converts the buildroot variable BR2_OPENPOWER_SIGNED_SECURITY_VERSION into a local environment variable in a specific spot of the sbe-p10 makefile.  This ensures that the sbe-p10 package is built with the same value as the rest of the build.

### DIFF
--- a/openpower/package/sbe-p10/sbe-p10.mk
+++ b/openpower/package/sbe-p10/sbe-p10.mk
@@ -30,7 +30,8 @@ endef
 
 define SBE_P10_INSTALL_IMAGES_CMDS
 	$(INSTALL) -D $(@D)/images/ipl_image_tool $(HOST_DIR)/usr/bin/
-	python2 $(@D)/src/build/sbeOpDistribute.py  --sbe_binary_dir=$(STAGING_DIR)/sbe_binaries --img_dir=$(@D)/images --sbe_binary_filename $(BINARY_SBE_FILENAME)
+	BR2_OPENPOWER_SIGNED_SECURITY_VERSION=${BR2_OPENPOWER_SIGNED_SECURITY_VERSION} \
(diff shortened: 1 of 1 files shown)


mabaiocchi:
@fspbld @dcrowell77 please review when you get the chance.

This is based off of the previously merged 1060 commit here: https://github.com/open-power/op-build/commit/2f064ce12d431c8c4b0cb9b12e7b7814b9ccab8c

openpower/package/sbe-p10/sbe-p10.mk
```diff
@@ -30,7 +30,8 @@ endef
 
 define SBE_P10_INSTALL_IMAGES_CMDS
 	$(INSTALL) -D $(@D)/images/ipl_image_tool $(HOST_DIR)/usr/bin/
-	python2 $(@D)/src/build/sbeOpDistribute.py  --sbe_binary_dir=$(STAGING_DIR)/sbe_binaries --img_dir=$(@D)/images --sbe_binary_filename $(BINARY_SBE_FILENAME)
+	BR2_OPENPOWER_SIGNED_SECURITY_VERSION=${BR2_OPENPOWER_SIGNED_SECURITY_VERSION} \
```
dcrowell77, line 33: This change doesn't make any sense to me so it must be some makefile magic?  You are setting an environment variable equal to a make variable?  (The line in a shell would be setting var to itself...)

mabaiocchi: Exactly.  I'm setting the environment variable BR2_OPENPOWER_SIGNED_SECURITY_VERSION that the SBE code is looking for to the buildroot variable "${BR2_OPENPOWER_SIGNED_SECURITY_VERSION}" (yes the same name).

The more complicated option is to update the SBE compile to take this value in as a parameter (which is what the openpower-pnor-p10 package does [HERE](https://github.com/open-power/op-build/blob/ibm-release-fw1060/openpower/package/openpower-pnor-p10/openpower-pnor-p10.mk).

dcrowell77 (approved):

PR #6365, opened 2025-05-07
Add support for common SBE and OCMBFW PNOR sections

dcrowell77:
This commit passes in common directories to build the SBE and OCMBFW PNOR sections/lids into the PNOR update_image.pl script.

These common directories can then be used to just build the SBE and OCMBFW one time and share with all of the system-specific directories.

This commit also picks up the latest main-p11 pnor commit to support these new directories.


dcrowell77:
Cherry-pick from working GHE

dcrowell77:
retest this please

PR #6376, opened 2025-05-09
Added consistent salt for HB HPT generation

matt-strom:
-Sets env var SALT_FOR_HASH_PAGE_TABLES so
 hostboot picks this up and is able to use the
 same value for each pass generating binaries


dcrowell77:
test this please

PR #6425, opened 2025-06-09
Buildroot 2025.02, kernel 5.10.237, skiboot 7.1+

rarbab:
Long overdue update to modern versions of buildroot, the kernel, and skiboot, for security and bug fixes.

On a RHEL8 builder, every defconfig in openpower/configs still builds for me.


runlevel5:
@rarbab thanks much for this work

I will let you know as soon as I am back from holiday

runlevel5:
> @runlevel5 Is there some reason we need 5.15? From a service perspective, the 5.10.x and 5.15.x LTS releases are both currently projected to have the same end-of-life.

@rarbab ultimately I want to bump to latest 6.15.x series because it has up-to-date firmwares for hardwares. For example  the RaptorCS Blackbird workstation would not be able to load up AMD graphic card driver at the early stage of boot menu.

runlevel5:
@rarbab is there any particular policy from IBM or OpenPOWER on choosing kernel version?

